PR #19, opened 2015-07-28
Allow 2.8.1 to be used

jsoucheiron:
This PR will let motor be installed with pymongo 2.8.1

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -123,7 +123,7 @@ setup(name='motor',
       install_requires=[
           'tornado >= 3.1',
           'greenlet >= 0.4.0',
-          'pymongo == 2.8.0',
+          'pymongo >= 2.8.0, <3',
       ],
       license='http://www.apache.org/licenses/LICENSE-2.0',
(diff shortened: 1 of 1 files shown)


ajdavis:
Hi, thanks for your interest! I'm not going to accept this -- you could *install* Motor with PyMongo 2.8.1 but Motor won't actually run:

https://github.com/mongodb/motor/blob/master/motor/__init__.py#L32

Motor requires PyMongo 2.8.0 exactly, for now. PyMongo 2.8.1 does include important bugfixes that I'd like Motor to benefit from, but I don't have time this quarter to complete a round of testing and development on Motor.

ajdavis:
That said, if there's a specific problem that the 2.8.0 dependency is causing for you, I could reconsider.

jsoucheiron:
It makes sense, thanks. I don't think we're affected by any of the bugs
fixed by 2.8.1.

2015-08-02 23:54 GMT+01:00 A. Jesse Jiryu Davis <notifications@github.com>:

PR #21, opened 2015-09-08
raise AutoReconnect instead of IncompleteReadError

jettify:
In case mongo sever goes down, StreamReader raises IncompleteReadError, when should just throw AutoReconnect.

In fix we just return ``b''`` so pymongo raises ``AutoReconnect`` for us, see:
https://github.com/mongodb/mongo-python-driver/blob/414250b5ef918656427e1a60fbf203bd023d543b/pymongo/network.py#L121-L131

@asvetlov @ajdavis  review please.


asvetlov:
LGTM. Thanks, @jettify

ajdavis:
Neat, thanks!

PR #24, opened 2015-11-28
added *callback* keyword to example

benediktschmitt:
Hi, the callback keyword was not present in the example.


ajdavis:
Thanks!

PR #25, opened 2016-01-24
Don't hardcode pymongo version in string

miedzinski:
As title says. Pretty much nothing changes.

### DIFF
--- a/motor/__init__.py
+++ b/motor/__init__.py
@@ -34,8 +34,9 @@ if pymongo.version != expected_pymongo_version:
     msg = (
         "Motor %s requires PyMongo at exactly version %s. "
         "You have PyMongo %s. "
-        "Do pip install pymongo==2.8.0"
-    ) % (version, expected_pymongo_version, pymongo.version)
+        "Do pip install pymongo==%s"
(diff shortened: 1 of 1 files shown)


ajdavis:
Do try to test your changes before submitting them to projects you'd like to contribute to =)

https://github.com/mongodb/motor/blob/master/motor/__init__.py#L32

Motor requires PyMongo 2.8.0 exactly because it is deeply coupled with PyMongo's internal implementation. I'd like to relax this requirement somewhat, but not yet

https://jira.mongodb.org/browse/MOTOR-87

miedzinski:
Did you read commit diff? It's just a small change, so no problems occur when someone edits expected_pymongo_version variable, but forgets to edit the string itself.
After merging pull request, the string will actually be different than current one (`pip install pymongo==2.8.0` -> `pip install pymongo==2.8`) - but it doesn't matter, doing `pip install pymongo==2.8` installs exactly 2.8, not 2.8.1, not 2.8rc1 or anything else. Same effect.

ajdavis:
Oh I see. Please accept my apology.

PR #26, opened 2016-03-07
Fix Python 3 map handling.

hellysmile:


ajdavis:
Interesting, thanks.

hellysmile:
Any chance to release this hotfix?

ajdavis:
Obviously https://pypi.python.org/pypi/motor/0.6.2

motor/motor_tornado.py
```diff
@@ -23,7 +23,7 @@
 from .metaprogramming import create_class_with_framework
 
 # See https://bugs.python.org/issue21720
-__all__ = map(str, ['MotorClient', 'MotorReplicaSetClient', 'Op'])
+__all__ = list(map(str, ['MotorClient', 'MotorReplicaSetClient', 'Op']))
```
drgarcia1986: Why not just?
```python
__all__ = ['MotorClient', 'MotorReplicaSetClient', 'Op']
```

ajdavis: See https://bugs.python.org/issue21720

PR #27, opened 2016-04-03
Remove redundancy code.

iceboy-sjtu:
Line 368-371 is duplicated with line 364-367.

Note that there is still problem that `_closed()` will always return `False` in `asyncio` framework because `AsyncioMotorSocket.fileno()` is not implemented.


ajdavis:
Thanks!

ajdavis:
I've filed the _closed bug, thanks for pointing that out: https://jira.mongodb.org/browse/MOTOR-107

PR #29, opened 2016-10-01
Drop distribute_setup.py

agronholm:
Distribute has been dead and buried for years – don't use it under any circumstances!


behackett:
It would be better to replace it with ez_setup.py. Given the python versions Motor supports, the latest version (28.0.0 as I write this) will work fine.

agronholm:
Alright, I'll make a new PR.

PR #30, opened 2016-10-01
Switch from distribute_setup.py to ez_setup.py

agronholm:
Distribute has been dead and buried for years. Setuptools should be used instead.


agronholm:
Do I need to add myself to the list of contributors for this?

behackett:
Thanks for the patch. I'm realizing we need to update the ez_setup.py in PyMongo master as well. It currently pulls setuptools 1.4.2 because PyMongo 2.x supports Python 2.4. PyMongo 3.x (github master) supports the same Python versions as Motor.

> Do I need to add myself to the list of contributors for this?

You can, but it's certainly not required.

agronholm:
When will PyMongo/Motor drop Python 2.6 support? After all it's been EOL'd for 3 years now.

ajdavis:
Motor supports a range of Tornado versions, and all the Python versions those Tornado versions work with.

Tornado just dropped Python 2.6 in Tornado 4.4 this July. Motor still supports Tornado back to version 3 ([MOTOR-83](https://jira.mongodb.org/browse/MOTOR-83)) because New Relic has [only begun to support Tornado 4](https://docs.newrelic.com/docs/agents/python-agent/hosting-mechanisms/introductory-tornado-4-support).

Supporting Python 2.6 doesn't cost me much and it's helpful to users. It's not my job to force them to upgrade Python.

agronholm:
Alright, if you're fine with it then I won't press the issue.

PR #31, opened 2016-10-11
Fix Python 3.5 aggregate example

jeanfrancisroy:
After wondering why Python 3.5 aggregate example was printing `None`s, I figured out that it had an extra line that was wrong.


ajdavis:
Thanks!

PR #34, opened 2017-04-23
Fix incorrect use of regexp flags

jwilk:
The regexp `sub()` method doesn't take flags as an argument.
The flags have to be supplied at compile time, and another part of code already does that.


ajdavis:
Thanks!

PR #36, opened 2017-09-15
allow AgnosticCursor.collation to be called as chained method

karolhor:
AgnosticCursor::collation can't be called as chained method

```python
db["test_collection"].find({}).sort("field").collation(Collation("pl")).to_list(length=1)
```

This PR resolves bug https://jira.mongodb.org/browse/MOTOR-154

### DIFF
--- a/test/tornado_tests/test_motor_cursor.py
+++ b/test/tornado_tests/test_motor_cursor.py
@@ -279,12 +279,15 @@ class MotorCursorTest(MotorMockServerTest):
 
     @gen_test
     def test_to_list_with_chained_collation(self):
+        if not (yield at_least(self.cx, (3, 4))):
+            raise SkipTest("collation requires MongoDB >= 3.4")
+
         yield self.make_test_data()
         cursor = self.collection.find({}, {'_id': 1}) \
             .sort([('_id', pymongo.ASCENDING)]) \
             .collation(Collation("en"))
         expected = [{'_id': i} for i in range(200)]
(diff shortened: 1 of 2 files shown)


test/tornado_tests/test_motor_cursor.py
```diff
@@ -277,6 +278,16 @@ def test_to_list_tailable(self):
             yield cursor.to_list(10)
 
     @gen_test
+    def test_to_list_with_chained_collation(self):
```
ajdavis: Please add something like:
```
        if not (yield at_least(self.cx, (3, 4))):
            raise SkipTest("collation requires MongoDB >= 3.4")
```

test/tornado_tests/test_motor_cursor.py
```diff
@@ -277,6 +278,16 @@ def test_to_list_tailable(self):
...
+        yield self.make_test_data()
+        cursor = self.collection.find({}, {'_id': 1}) \
+            .sort([('_id', pymongo.ASCENDING)]) \
+            .collation(Collation("en"))
+        expected = [{'_id': i} for i in range(200)]
+        (result, error), _ = yield gen.Task(cursor.to_list, length=1000)
```
ajdavis: This can just be:
```
result = yield cursor.to_list(length=1000)
```
No need for gen.Task in the last few years.

karolhor:
@ajdavis thx for review. I've updated PR with your suggestions

ajdavis:
Thanks!

PR #37, opened 2017-12-13
Fix causal consistency example.

ShaneHarvey:
By default, operations are only casually consistent within the same session. The `insert` must use the same session as the `find`.


ajdavis:
Ooh, thank you. I'll fix my blog post too.

ShaneHarvey:
Cheers!

PR #38, opened 2018-03-07
Add optional dependencies, forwarding them to pymongo

arthurdarcet:
`pymongo` supports three optional dependencies, it would be nice to support them as well here, so that we can depend directly on `motor[srv]` instead of having to list both `motor` and `pymongo[srv]`

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -160,6 +160,10 @@ setup(name='motor',
       author_email='jesse@mongodb.com',
       url='https://github.com/mongodb/motor/',
       install_requires=install_requires,
+      extras_require={
+        key: 'pymongo[{}]'.format(key)
+        for key in ('tls', 'srv', 'gssapi')
+      },
       license='http://www.apache.org/licenses/LICENSE-2.0',
       classifiers=[c for c in classifiers.split('\n') if c],
       keywords=[


eguven:
Not sure if this would be the same behaviour using pypi but it doesn't seem to work locally.

```shell
$ pip install -U -e .[srv]
Obtaining file:///***/***/motor
  motor 1.3.dev0 does not provide the extra 'srv'
Collecting pymongo<4,>=3.4 (from motor==1.3.dev0)
  Using cached pymongo-3.6.1-cp35-cp35m-macosx_10_6_intel.whl
Installing collected packages: pymongo, motor
  Running setup.py develop for motor
Successfully installed motor pymongo-3.6.1
$ pip freeze
-e git+git@github.com:eguven/motor.git@1a5055d06391e779f27ca6cc1e8c7ea01c59baad#egg=motor
pymongo==3.6.1
```

arthurdarcet:
@eguven there was indeed a typo in my commit, which I fixed.
But it's still not working: the optional dependency are ignored, probably because pymongo is listed in both install_requires and extras_require. It looks like a pip bug to me, but I don't have time to look into it for now, I'll open an issue with them when I have the time

arthurdarcet:
Actually, the issue already exists on pip: https://github.com/pypa/pip/pull/3878
This PR should work once pypa/pip#3878 is merged

ajdavis:
OK, if this feature doesn't work with pip today, it will be a long time before we can rely on all pip users to have upgraded to a version of pip that supports it. A few years minimum. Thanks for your work, unfortunately I can't accept this PR.

PR #42, opened 2018-09-25
fix: fix sample document

fanjindong:
I found a sample error on the motor.readthedocs.io, then I fixed the sample document


ajdavis:
Thanks!

PR #43, opened 2018-09-29
Regenerate test certificates

onovy:
Old one is not supported in OpenSSL 1.1.1


onovy:
Old one uses SHA1 hash, which is disabled in OpenSSL 1.1.1. New one uses SHA256.

Sorry, I don't have commands in my history, but I created new self-signed CA and generated cert from it. I kept all information from old certificates.

ajdavis:
I see, thanks for the information! I've decided to just delete Motor's test certificates instead, and use the certificates that all MongoDB drivers share for testing:

https://github.com/mongodb-labs/drivers-evergreen-tools/tree/master/.evergreen/x509gen

behackett:
Interesting. The certs we use for all the other drivers could have this problem too. I'll take a look.

ajdavis:
Oh good point. Yes they do use SHA1 and need to be regenerated.

ajdavis:
@onovy I agree with you that we should regenerate these certs with SHA256 signatures. However, I can't find any documentation stating that SHA1 signatures are prohibited by OpenSSL 1.1.1. Can you tell me why you think SHA1 is prohibited now?

onovy:
@ajdavis After more investigation I found that OpenSSL 1.1.1 didn't prohibited SHA1, but default configuration in Debian did. So not all OpenSSL 1.1.1 installs are affected, but all Debians and Debian-based (Ubuntu, etc.). Sorry for wrong information.

ajdavis:
OK great, thanks! So, we should regenerate the certs, but this doesn't affect all OpenSSL 1.1.1 users.

onovy:
right, thanks.

ajdavis:
Tracking here: https://jira.mongodb.org/browse/DRIVERS-575

PR #45, opened 2019-03-01
Align requirements in code to requirements.rst

crusaderky:
This is based on the very strong assumption that requirements.rst is the authoritative source of truth...


ajdavis:
Thanks for the patch! I'm not sure we should accept this, since it's a fair amount of logic and a bug (now or in the future) in this logic would prevent installation. I think the standard and safe way to express requirements in setup.py is in install_requires and python_requires. I propose that you *only* update the contents of those two variables, and remove all additional code to check Python and Tornado versions.

crusaderky:
@ajdavis changed as requested

setup.py
```diff
@@ -30,13 +30,20 @@
 
 long_description = open("README.rst").read()
 
-install_requires = ['pymongo>=3.6,<4']
+install_requires = ['pymongo>=3.7']
 
 tests_require = ['mockupdb>=1.4.0']
 
 if sys.version_info[0] < 3:
     install_requires.append('futures')
```
ShaneHarvey: This kind of version checking does not work in setup.py does not actually work because motor uploads a single py2/py3 universal wheel to pypi: https://pypi.org/project/motor/#files.

If the *universal wheel is built* with python 2 then `install_requires` will include `futures`, but if it's built with python 3 then it will not include `futures`. As proof that this doesn't work let's install motor on python 2.7:
```
(venv) ➜  tmp-motor pip install motor
DEPRECATION: Python 2.7 will reach the end of its life on January 1st, 2020. Please upgrade your Python as Python 2.7 won't be maintained after that date. A future version of pip will drop support for Python 2.7.
Collecting motor
  Using cached https://files.pythonhosted.org/packages/28/ca/a2e740ad2822cbf0d2ac4a3b2b48e043142014ab5c8c2129fa5c71c8cdb5/motor-2.0.0-py2.py3-none-any.whl
Collecting pymongo<4,>=3.6 (from motor)
  Using cached https://files.pythonhosted.org/packages/81/ed/ab60269689de46be0964fd93494f9f86a5196802a9156eddc5e5411750c9/pymongo-3.7.2-cp27-cp27m-macosx_10_13_intel.whl
Installing collected packages: pymongo, motor
Successfully installed motor-2.0.0 pymongo-3.7.2
(venv) ➜  tmp-motor pip freeze
DEPRECATION: Python 2.7 will reach the end of its life on January 1st, 2020. Please upgrade your Python as Python 2.7 won't be maintained after that date. A future version of pip will drop support for Python 2.7.
motor==2.0.0
pymongo==3.7.2
```

There are alternative ways to specify python version specific dependencies but they may or may not work depending on the user's version setuptools or pip: https://hynek.me/articles/conditional-python-dependencies/

I suggest we stop publishing universal wheels and upload one for each supported python version. This is outside the scope of this PR so I'll file a new issue for it.

setup.py
```diff
@@ -30,13 +30,20 @@
...
 
+# Optional dependency with a minimum version
+try:
+    import tornado
+    install_requires.append('tornado>=4.0')
+except ImportError:
+    pass
```
ShaneHarvey: Why do we only include tornado as a dependency when the build environment includes tornado? See my comment above for why I don't think this will work as expected for wheels -- universal or otherwise.

crusaderky:
What's the point of a wheel? This is a pure python package...

ShaneHarvey:
Wheels have many benefits even for pure python packages: 

1. Faster installation for **pure Python** and native C extension packages.
1. Avoids arbitrary code execution for installation. (**Avoids setup.py**)
1. Installation of a C extension does not require a compiler on Linux, Windows or macOS.
1. Allows better caching for testing and continuous integration.
1. Creates .pyc files as part of installation to ensure they match the Python interpreter used.
1. More consistent installs across platforms and machines.

https://pythonwheels.com/

ajdavis:
Thanks for reviewing this, Shane! Good points. I've pushed another commit to this branch (thanks @crusaderky for giving permission to add commits to this pull request). My new commit reverts these changes except for a few things I think are useful:

1.  Really update the PyMongo requirement to 3.7, which I'd left incomplete in [MOTOR-225](https://jira.mongodb.org/browse/MOTOR-225)
2. Keep the "python_requires" argument that @crusaderky added to setuptools.setup().

Shane, I notice that once upon a time in motor/\_\_init\_\_.py I had written logic to double-check the PyMongo version after importing PyMongo. I've fixed that code to check for the proper PyMongo version now (it had still been checking for "pymongo>=3.4"), but I think it should just be deleted. Leaving that to your judgment.

crusaderky:
Sorry, this fell through the cracks.
Updated and ready for merge now.

setup.py
```diff
@@ -157,6 +157,8 @@ def run(self):
       author='A. Jesse Jiryu Davis',
       author_email='jesse@mongodb.com',
       url='https://github.com/mongodb/motor/',
+      python_requires='>=2.7, !=3.0.*, !=3.1.*, !=3.2.*, != 3.3.*, '
+                      '!=3.5.0, !=3.5.1',
```
ShaneHarvey: Can you add `'!=3.4.*'` to the list and wrap the multi-line string in parentheses `('...')`?

crusaderky: Reformatted.
3.5 is supported according to requirements.rst...

ShaneHarvey: Oh you're totally right!

ShaneHarvey:
Thanks @crusaderky!

PR #47, opened 2019-07-15
Add save command

levsh:
https://docs.mongodb.com/manual/reference/method/db.collection.save/


behackett:
Thanks for contributing. The save method in PyMongo is deprecated, so we won't add support for it to Motor.

PR #59, opened 2020-07-08
Replace usages of self.__class__ while calling functions from super class with appropriate class names WIP

kirantambe:
Explicitly specified AgnosticBaseProperties while calling super in AgnosticDatabase and AgnosticCollection so that no errors are raised while extending these classes

### DIFF
--- a/motor/core.py
+++ b/motor/core.py
@@ -448,7 +448,7 @@ class AgnosticDatabase(AgnosticBaseProperties):
         delegate = kwargs.get('_delegate') or Database(
             client.delegate, name, **kwargs)
 
-        super(self.__class__, self).__init__(delegate)
+        super(AgnosticBaseProperties, self).__init__(delegate)
 
     def aggregate(self, pipeline, **kwargs):
         """Execute an aggregation pipeline on this database.
@@ -676,7 +676,7 @@ class AgnosticCollection(AgnosticBaseProperties):
             read_preference=read_preference, write_concern=write_concern,
             read_concern=read_concern)
 
-        super(self.__class__, self).__init__(delegate)
+        super(AgnosticBaseProperties, self).__init__(delegate)
         self.database = database
 
     def __getattr__(self, name):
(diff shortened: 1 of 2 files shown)


motor/core.py
```diff
@@ -448,7 +448,7 @@ def __init__(self, client, name, **kwargs):
         delegate = kwargs.get('_delegate') or Database(
             client.delegate, name, **kwargs)
 
-        super(self.__class__, self).__init__(delegate)
+        super(AgnosticBaseProperties, self).__init__(delegate)
```
ShaneHarvey, line 451: This should be `AgnosticDatabase` instead of `AgnosticBaseProperties`.

motor/core.py
```diff
@@ -676,7 +676,7 @@ def __init__(self, database, name, codec_options=None,
             read_preference=read_preference, write_concern=write_concern,
             read_concern=read_concern)
 
-        super(self.__class__, self).__init__(delegate)
+        super(AgnosticBaseProperties, self).__init__(delegate)
```
ShaneHarvey, line 679: This should be `AgnosticCollection` instead of `AgnosticBaseProperties`.

motor/core.py
kirantambe: This does not work because at run time self is an instance of `motor.motor_asyncio.AsyncIOMotorDatabase` which is not a child class of `AgnosticDatabase`, hence the previous use of `self.__class__`
I don't see a problem using `super(AgnosticBaseProperties, self).__init__(delegate)` since thats exactly what we are trying to invoke here, right?

ShaneHarvey: Yes that's a great point. I understand the problem now. Still, using the parent class in these super calls worries me a bit because the usual pattern is to use the current class.  We're currently dropping support for Python 2 (https://github.com/mongodb/motor/pull/61). After dropping 2, I think it would be possible to use "bare" super() calls everywhere. 

Could you try using `super()` instead and ignore the Python 2 test failures? It would also be great if you could add a test for your use-case of extending these classes.

prashantmital:
Done in #89. Thank you @kirantambe for bringing this to our attention.

PR #61, opened 2020-07-24
MOTOR-419 Drop support for Python versions older than 3.5.2

prashantmital:


setup.py
```diff
@@ -33,14 +25,10 @@
 
 long_description = open("README.rst").read()
 
-install_requires = ['pymongo>=3.10,<4']
+install_requires = ['pymongo>3.10,<4']
 
 tests_require = ['mockupdb>=1.4.0']
```
ShaneHarvey, line 31: I think we should add a guard to setup.py like this:
```
if sys.version_info[:2] < (3, 5):
    raise Exception("This version of Motor requires Python >=3.5")
```

This will cause installation to fail early on clients which don't support `python_requires`.

doc/requirements.rst
```diff
@@ -56,35 +45,39 @@ Motor and PyMongo
...
++---------------+-----+-----+-----+-----+-----+-----+-----+-----+-----+-----+-----+
+|               | 1.2 |**N**|**N**|  Y  |  Y  |  Y  |  Y  |  Y  |**N**|**N**|**N**|
++---------------+-----+-----+-----+-----+-----+-----+-----+-----+-----+-----+-----+
+|               | 1.3 |**N**|**N**|  Y  |  Y  |  Y  |  Y  |  Y  |**N**|**N**|**N**|
++---------------+-----+-----+-----+-----+-----+-----+-----+-----+-----+-----+-----+
+|               | 2.0 |**N**|**N**|**N**|  Y  |  Y  |  Y  |  Y  |  Y  |  Y  |**N**|
++---------------+-----+-----+-----+-----+-----+-----+-----+-----+-----+-----+-----+
+|               | 2.1 |**N**|**N**|**N**|  Y  |  Y  |  Y  |  Y  |  Y  |  Y  |**N**|
++---------------+-----+-----+-----+-----+-----+-----+-----+-----+-----+-----+-----+
+|               | 2.2 |**N**|**N**|**N**|  Y  |  Y  |  Y  |  Y  |  Y  |  Y  |  Y  |
```
ShaneHarvey, line 71: Pymongo still supports MongoDB 2.6 so I think motor (2.X) should still support it as well.

doc/requirements.rst
```diff
@@ -118,8 +113,7 @@ the "async for" protocol implemented in Python 3.5.0 and 3.5.1. Motor continues
 to work with "async for" loops in Python 3.5.2 and later.
 
 Motor 1.2.5 and 1.3.1 add compatibility with Python 3.7, but at the cost of
-dropping Python 3.4.3 and older. Python 3.4.4 and later are supported by all
-Motor versions.
+dropping Python 3.4.3 and older.
```
ShaneHarvey, line 116: Let's add a line saying that Motor 2.2 dropped support for python 2.7 and 3.4.

doc/requirements.rst
prashantmital: I believe officially support was dropped after we stopped testing Motor explicitly against MongoDB 2.6. While I agree with your rationale, I am not particularly excited about re-introducing 2.6 support after 2 versions have already advertised it as being unsupported. Maybe @techbelle can weigh in here with some Motor usage numbers by MongoDB version - I suspect usage skews heavily towards newer server versions for this driver.

setup.py
prashantmital: Sounds good. Done.

doc/requirements.rst
prashantmital: Done.

doc/requirements.rst
ShaneHarvey: 2.6 was dropped in https://jira.mongodb.org/browse/MOTOR-252.

PR #65, opened 2020-07-29
MOTOR-587 Remove py2 compatibility submodule

prashantmital:


test/tornado_tests/test_motor_replica_set.py
```diff
@@ -66,8 +64,8 @@ def test_auth_network_error(self):
         credentials = pymongo.auth._build_credentials_tuple(
```
ShaneHarvey, line 64: Can you look into the history of this test? I suspect it came from a time when motor would customize PyMongo's internal Pool and Monitor classes. If so we can delete this test altogether.

prashantmital: So the test first appeared in 0.5: https://github.com/mongodb/motor/tree/0.5

Back then it does seem like we did some patching of PyMongo's internals. AFAICT all tests in this file and its counterpart in the `asyncio` tests are redundant - i.e. they are tested in `test_<motor/asyncio>_client.py` as well so we should just remove these test files entirely. Broadly, these tests are from a time when the RS client was a different class.

prashantmital: I am removing this test in this PR. And I have opened https://jira.mongodb.org/browse/MOTOR-588 to deal with the remaining tests in these files.

PR #66, opened 2020-07-31
MOTOR-488 Don't call set_result on a cancelled tasks

ShaneHarvey:
This change prevents Motor from failing with `asyncio.exceptions.InvalidStateError: invalid state` when tasks are cancelled by the application.

### DIFF
--- a/motor/core.py
+++ b/motor/core.py
@@ -1538,6 +1538,9 @@ class _LatentCursor(object):
     _CommandCursor__killed = False
     cursor_id = None
 
+    def _CommandCursor__end_session(self, *args, **kwargs):
+        pass
+
     def _CommandCursor__die(self, *args, **kwargs):
         pass
 
--- a/motor/core.py
+++ b/motor/core.py
@@ -1356,6 +1356,9 @@ class AgnosticBaseCursor(AgnosticBase):
         return future
 
     def _to_list(self, length, the_list, future, get_more_result):
+        # Return early if the task was cancelled.
+        if future.done():
+            return
         # get_more_result is the result of self._get_more().
         # to_list_future will be the result of the user's to_list() call.
         try:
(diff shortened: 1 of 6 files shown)


motor/core.py
```diff
@@ -1535,6 +1537,9 @@ class _LatentCursor(object):
     _CommandCursor__killed = False
     cursor_id = None
 
+    def _CommandCursor__end_session(self, *args, **kwargs):
```
ShaneHarvey, line 1540: This was required to fix this error:
```
ERROR: test_cancelled_error_fetch_next_aggregate (test.asyncio_tests.test_asyncio_cursor.TestAsyncIOCursor)
  File "/Users/shane/git/motor/test/asyncio_tests/__init__.py", line 48, in __call__
    result = self.orig_method()
  File "/Users/shane/git/motor/test/test_environment.py", line 283, in wrap
    return f(*args, **kwargs)
  File "/Users/shane/git/motor/test/asyncio_tests/__init__.py", line 217, in wrapped
    raise coro_exc from None
  File "/Library/Frameworks/Python.framework/Versions/3.8/lib/python3.8/concurrent/futures/thread.py", line 57, in run
  File "/Users/shane/git/mongo-python-driver/pymongo/command_cursor.py", line 87, in __die
    print(self.__address)
AttributeError: '_LatentCursor' object has no attribute '_CommandCursor__address'
```

ShaneHarvey:
evergreen retry

ShaneHarvey:
evergreen retry

ShaneHarvey:
Note: the failures are caused by the release of pymongo 3.11 and the current lack of allowDiskUse support (https://jira.mongodb.org/browse/MOTOR-454).

motor/core.py
prashantmital: Sorry, what is this failure? I can't see this print statement in the code. I can see that `__die` calls `_CommandCursor__end_session` so it makes sense that you needed this method but I can't figure out what this traceback is telling us.

ShaneHarvey: Updated to include the correct error traceback. The earlier print-statement one was from debugging. I also fixed the `KeyError: 'exception'` issue. The context's exception field is optional: https://docs.python.org/3/library/asyncio-eventloop.html#asyncio.loop.call_exception_handler

PR #75, opened 2020-08-05
MOTOR-538 Revise issues and help sections of documentation

prashantmital:
I made changes to make the Readme and documentation index page consistent with that of PyMongo - I think this will make it more maintainable in the long run. If you don't share this opinion, I can make surgical changes and leave in place the old language and structure.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -47,10 +47,10 @@ Think you’ve found a bug? Want to see a new feature in Motor? Please open a
 case in our issue management tool, JIRA:
 
 - `Create an account and login <https://jira.mongodb.org>`_.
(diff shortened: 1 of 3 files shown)


README.rst
```diff
@@ -75,18 +77,49 @@ Please include all of the following information:
...
+Installation
+============
+
+Motor can be installed with `pip <http://pypi.python.org/pypi/pip>`_::
+
+  $ pip install motor
+
+You can also download the project source and do::
+
+  $ python setup.py install
```
ShaneHarvey, line 96: Let's remove this option. Or replace it with `python3 -m pip install .`

prashantmital: Removed it from the README. Added to the Installation section of docs.

PR #76, opened 2020-08-06
MOTOR-320 Create ChangeStream cursor in async with __aenter__

ShaneHarvey:
This change makes motor's ChangeStream work more like PyMongo's by establishing the server-side cursor earlier. PyMongo creates the cursor in the call to `watch()`. To do that in Motor we would need to make `watch` async  (`def watch(...)` -> `async def watch(...)`) which is a backward breaking change. Instead, Prashant had the clever idea to create the cursor in `__aenter__`. This will resolve the user's complaint in https://jira.mongodb.org/browse/MOTOR-320.

### DIFF
--- a/motor/core.py
+++ b/motor/core.py
@@ -1758,11 +1758,14 @@ class AgnosticChangeStream(AgnosticBase):
     __anext__ = next
 
     async def __aenter__(self):
+        if not self.delegate:
+            loop = self.get_io_loop()
+            await self._framework.run_on_executor(loop, self._lazy_init)
         return self
 
     async def __aexit__(self, exc_type, exc_val, exc_tb):
         if self.delegate:
-            self.delegate.close()
+            await self.close()
 
     def get_io_loop(self):
         return self._target.get_io_loop()
--- a/motor/core.py
+++ b/motor/core.py
@@ -1666,6 +1666,9 @@ class AgnosticChangeStream(AgnosticBase):
         self._lazy_init()
         return self.delegate.try_next()
 
+    async def _run_on_executor(self, func):
+        return await self._framework.run_on_executor(self.get_io_loop(), func)
+
     async def next(self):
         """Advance the cursor.
 
@@ -1704,8 +1707,7 @@ class AgnosticChangeStream(AgnosticBase):
         example above, you can also iterate the change stream by calling
         ``await change_stream.next()`` repeatedly.
         """
-        loop = self.get_io_loop()
-        return await self._framework.run_on_executor(loop, self._next)
+        return await self._run_on_executor(self._next)
 
     async def try_next(self):
         """Advance the cursor without blocking indefinitely.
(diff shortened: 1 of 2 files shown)


motor/core.py
```diff
@@ -1758,11 +1758,14 @@ def __aiter__(self):
...
     async def __aenter__(self):
+        if not self.delegate:
+            loop = self.get_io_loop()
+            await self._framework.run_on_executor(loop, self._lazy_init)
         return self
 
     async def __aexit__(self, exc_type, exc_val, exc_tb):
         if self.delegate:
-            self.delegate.close()
+            await self.close()
```
prashantmital, line 1768: Nicely spotted optimization.

motor/core.py
```diff
@@ -1758,11 +1758,14 @@ def __aiter__(self):
     __anext__ = next
 
     async def __aenter__(self):
+        if not self.delegate:
```
prashantmital, line 1761: Can we do a refactor in this class such that `_lazy_init`, `_next` and `_try_next` run their logic on an executor by default. It's rather odd that we have to manually do that part each time. For example, for lazy init this would look like:
```
    async def _lazy_init(self):
        if not self.delegate:
            loop = self.get_io_loop()
            self.delegate = await self._framework.run_on_executor(
                loop, functools.partial(
                    self._target.delegate.watch, **self._kwargs))
```

This would, in turn, change this `__aenter__` method as follows:

```
async def __aenter__(self):
    await self._lazy_init()
    return self
```

prashantmital:
For some reason this breaks a test on 4.4 sharded

```
======================================================================
 FAIL [1.131s]: test_async_with_creates_cursor (asyncio_tests.test_asyncio_change_stream.TestAsyncIOChangeStream)
   File "/data/mci/7ebbf9fb06788935d797dbf4c5dab33a/src/test/asyncio_tests/__init__.py", line 48, in __call__
     result = self.orig_method()
   File "/data/mci/7ebbf9fb06788935d797dbf4c5dab33a/src/test/asyncio_tests/__init__.py", line 212, in wrapped
     self.loop.run_until_complete(task)
   File "/opt/python/3.8/lib/python3.8/asyncio/base_events.py", line 616, in run_until_complete
     return future.result()
   File "/opt/python/3.8/lib/python3.8/asyncio/tasks.py", line 483, in wait_for
     return fut.result()
   File "/data/mci/7ebbf9fb06788935d797dbf4c5dab33a/src/test/asyncio_tests/test_asyncio_change_stream.py", line 222, in test_async_with_creates_cursor
```

motor/core.py
ShaneHarvey: I refactored to add a helper method which runs `run_on_executor`  instead of making the change you suggested. We need `_lazy_init` to be synchronous because it's called in the synchrotests: https://github.com/mongodb/motor/blob/8208420d613bfe1de0dfee55baf54c62b3b25564/synchro/__init__.py#L139-L140

ShaneHarvey:
Reverted the `run_on_executor` because I suspect it causes the test failures.

PR #78, opened 2020-08-06
MOTOR-597 Stop using deprecated loop arguments and other improvements

ShaneHarvey:
Jira: https://jira.mongodb.org/browse/MOTOR-597

### DIFF
--- a/motor/frameworks/asyncio/__init__.py
+++ b/motor/frameworks/asyncio/__init__.py
@@ -97,7 +97,7 @@ def chain_return_value(future, loop, return_value):
         else:
             chained.set_result(return_value)
 
-    future.add_done_callback(functools.partial(loop.add_callback, copy))
+    future._future.add_done_callback(functools.partial(loop.add_callback, copy))
     return chained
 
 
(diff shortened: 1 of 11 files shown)


motor/frameworks/asyncio/__init__.py
```diff
@@ -97,12 +97,12 @@ def copy(_future):
         else:
             chained.set_result(return_value)
 
-    future._future.add_done_callback(functools.partial(loop.add_callback, copy))
```
prashantmital, line 100: I had been a bit confused by this back when I was working on the PR to drop older Pythons. can you clarify what is `future._future` and why we don't need it anymore?

ShaneHarvey: This was and still is totally broken and completely untested. I'll open a new ticket for it.

prashantmital: IIRC `to_list` breaks without this method. I tried to remove it when I was updating the framework files but couldn't figure out a way to get rid of the damn thing.

ShaneHarvey: I'll fix this in https://jira.mongodb.org/browse/MOTOR-600.

PR #80, opened 2020-08-07
MOTOR-582 Deprecate Cursor.fetch_next and Cursor.next_object

prashantmital:


motor/core.py
```diff
@@ -1139,11 +1140,15 @@ def _get_more(self):
     @property
     @coroutine_annotation
     def fetch_next(self):
```
ShaneHarvey, line 1142: We still use fetch_next and next_object in `__anext__`.

motor/core.py
```diff
@@ -1120,11 +1119,19 @@ def __init__(self, cursor, collection):
...
-        # An optimization: skip the "await" if possible.
-        if self._buffer_size() or await self.fetch_next:
-            return self.next_object()
-        raise StopAsyncIteration()
+    async def next(self):
+        """Advance the cursor.
+
+        .. versionadded:: 2.2
+        """
+        if not self.alive:
```
prashantmital, line 1127: We have to explicitly check this condition here (instead of relying on the same exception to be raised by `_get_more` and catching it here) because using `except InvalidOperation: raise StopAsyncIteration` actually masks command errors that occur during the `getMore`.

motor/core.py
```diff
@@ -1254,38 +1280,21 @@ def each(self, callback):
...
+                async for doc in self:
+                    # The callback closed this cursor?
+                    if self.closed:
+                        return
+                    # Quit if callback returns exactly False (not None). Note we
+                    # don't close the cursor: user may want to resume iteration.
+                    if callback(doc, None) is False:
+                        return
+                callback(None, None)
+            except Exception as exc:
```
prashantmital, line 1297: This is really weird design but we have to stick to it in the interest of maintaining backwards-compatibility. In 3.0 we can just raise the error from this coroutine instead of relying on the user-supplied callback to do so.

motor/core.py
```diff
@@ -1342,60 +1356,29 @@ def to_list(self, length):
...
 
     def get_io_loop(self):
         return self.collection.get_io_loop()
 
     async def close(self):
-        """Explicitly kill this cursor on the server. Call like (in Tornado):
-
-        .. code-block:: python
+        """Explicitly kill this cursor on the server.
+        Call like::
```
prashantmital, line 1378: Without this change, doctest complained as follows:

```
/Users/pmital/Developer/motor/motor/motor_tornado.py:docstring of motor.motor_tornado.MotorCursor.close:3: WARNING: Literal block expected; none found.
/Users/pmital/Developer/motor/motor/motor_tornado.py:docstring of motor.motor_tornado.MotorCommandCursor.close:3: WARNING: Literal block expected; none found.
```

doc/developer-guide.rst
```diff
@@ -32,6 +32,7 @@ A Motor framework is a module implementing these properties and functions:
 - ``coroutine`` (**DEPRECATED**)
 - ``get_event_loop``
 - ``get_future``
+- ``create_task``
```
prashantmital, line 35: This can be added because https://www.tornadoweb.org/en/branch5.1/_modules/tornado/platform/asyncio.html#AsyncIOMainLoop offers access to the underlying `asyncio` event loop and starting in this version we only support Python and Tornado versions where this API will work.

motor/core.py
```diff
@@ -1342,60 +1356,29 @@ def to_list(self, length):
...
+                return the_list
+            # Else
+            async for doc in self:
+                the_list.append(doc)
+                if (length is not None and
+                        len(the_list) >= length):
+                    break
+            return the_list
+
+        return self._framework.create_task(self.get_io_loop(), _coro())
```
prashantmital, line 1371: I think this pattern is a great example of why we should clean up the codebase to stop returning futures. The code is much more concise, easy to read, and easy to reason about.

doc/api-tornado/cursors.rst
```diff
@@ -5,6 +5,7 @@
 
 .. autoclass:: MotorCursor
   :members:
+  :inherited-members:
```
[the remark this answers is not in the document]
ShaneHarvey: Nice find!

doc/examples/tailable-cursors.rst
```diff
@@ -3,27 +3,43 @@
 Motor Tailable Cursor Example
 =============================
 
-This example describes using Motor with Tornado. Beginning in
-version 0.5 Motor can also integrate with asyncio instead of Tornado.
+By default, MongoDB will automatically close a cursor when the client has
```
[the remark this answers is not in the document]
ShaneHarvey: ack.

motor/core.py
```diff
@@ -1139,53 +1146,63 @@ def _get_more(self):
...
-          >>> IOLoop.current().run_sync(f)
-          0, 1, 2, 3, 4, done
+        .. versionchanged:: 2.2
+           Deprecated.
 
         .. _`large batches`: https://docs.mongodb.com/manual/tutorial/iterate-a-cursor/#cursor-batches
+        .. _`gen.coroutine`: http://tornadoweb.org/en/stable/gen.html
         """
+        warnings.warn("The fetch_next property is deprecated and will be "
+                      "removed in Motor 3.0. Use Use `async for` to iterate "
```
ShaneHarvey, line 1202: "Use Use" -> "Use"

motor/core.py
ShaneHarvey: Ack. Minor nit: We could reduce this code to:
```python
        if self.alive and (self._buffer_size() or await self._get_more()):
            return next(self.delegate)
        raise StopAsyncIteration
```

motor/core.py
```diff
@@ -1120,11 +1119,19 @@ def __init__(self, cursor, collection):
     def __aiter__(self):
         return self
 
-    async def __anext__(self):
-        # An optimization: skip the "await" if possible.
-        if self._buffer_size() or await self.fetch_next:
-            return self.next_object()
-        raise StopAsyncIteration()
+    async def next(self):
```
ShaneHarvey, line 1122: Why add `next`?

motor/core.py
ShaneHarvey: Can you add a blank line before "Call like:"?

test/tornado_tests/test_motor_core.py
```diff
@@ -80,7 +80,7 @@ def attrs(klass):
 
 pymongo_cursor_only = set([
     'count',
-    'retrieved']).union(pymongo_only)
```
ShaneHarvey, line 83: I don't think this is quite right. We should keep the union here and adjust `pymongo_only` to be correct (assuming we want to keep the newly added `next()`).

motor/core.py
prashantmital: It allows users to still iterate change streams in a while loop like so:

```
while cursor.alive:
    doc = await cursor.next()
```

Its not a particularly compelling reason, I know but I don't see what harm it does to include it.

motor/core.py
prashantmital: Nice. Done.

motor/core.py
prashantmital: Done.

motor/core.py
prashantmital: Done.

test/tornado_tests/test_motor_core.py
prashantmital: We can't really do that since `pymongo_only` is also used to compare attrs of `collection` and `database` objects - both of which have a `next` method in PyMongo but not in Motor.

prashantmital: On a closer look, it is unclear to me why those classes have `next` in PyMongo - they simply raise `TypeError` - but that isn't something that can be litigated on this PR.

ShaneHarvey: Ack.

motor/core.py
ShaneHarvey: I guess change stream cursors already have `next()` and `__anext__` too so I guess this is fine. Can you call out the new method in the changelog?

prashantmital: Added the changelog entry.

PR #81, opened 2020-08-10
MOTOR-598 Fix doctests and vendor docstrings with code examples

prashantmital:
Also addresses MOTOR-599

### DIFF
--- a/motor/docstrings.py
+++ b/motor/docstrings.py
@@ -1053,8 +1053,7 @@ Pass a field name and a direction, either
 
 .. doctest:: sort
 
-  >>> @gen.coroutine
-  ... def f():
+  >>> async def f():
   ...     cursor = collection.find().sort('_id', pymongo.DESCENDING)
   ...     docs = await cursor.to_list(None)
   ...     print([d['_id'] for d in docs])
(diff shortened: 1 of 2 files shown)


motor/docstrings.py
```diff
@@ -1186,3 +1182,79 @@ async def coro():
...
+
+.. versionchanged:: 2.2
+   Deprecated.
+
+.. _fsyncUnlock command: https://docs.mongodb.com/manual/reference/command/fsyncUnlock/
+"""
+
+reindex_doc = """**DEPRECATED**: Rebuild all indexes on this collection.
+
+Deprecated. Use :meth:`~motor.motor_tornado.MotorDatabase.command`
```
ShaneHarvey, line 1211: This reminds me of the problem noticed in https://jira.mongodb.org/browse/MOTOR-560. Any bright ideas for how to make this point to `motor.motor_tornado.MotorDatabase.command` OR `motor.motor_asyncio.AsyncIOMotorDatabase.command` depending on which API is being viewed?

prashantmital: I looked into this and wasn't able to figure a way out. I think our best bet to resolve this is in 3.0 when we move to a framework-agnostic API.

PR #84, opened 2020-08-12
MOTOR-596 Deprecate fsync helper

prashantmital:


ShaneHarvey:
What does the deprecation warning look like to the user in this case?

prashantmital:
The interpreter needs to be run with `-W default` (or equivalent) to see warnings produced on thread pool executor threads. 

Test script:
```
$ cat sample.py 
import asyncio


async def main():
    client = AsyncIOMotorClient()
    print(await client.admin.command('ping'))
    print(client.nodes)
    await client.fsync()


asyncio.run(main())
```

Running it + output:
```
$ python -Wdefault sample.py
{'ok': 1.0}
frozenset({('localhost', 27017)})
/Users/pmital/.pyenv/versions/3.7.5/lib/python3.7/concurrent/futures/thread.py:57: DeprecationWarning: fsync is deprecated. Use client.admin.command('fsync') instead.
```

Running it without `-Wdefault`:
```
$ python sample.py
{'ok': 1.0}
frozenset({('localhost', 27017)})
```

I suppose w could add a warning directly in Motor to work around the need for the extra flag but I think most users will need to run using `-Werror` or similar during migration so this isn't really a big deal?

prashantmital:
Opened https://jira.mongodb.org/browse/MOTOR-607

PR #85, opened 2020-08-12
MOTOR-420 Remove legacy APIs/patterns from documentation

prashantmital:
This will be easier to review after we merge #80

### DIFF
--- a/doc/api-asyncio/cursors.rst
+++ b/doc/api-asyncio/cursors.rst
@@ -7,7 +7,6 @@
   :members:
   :inherited-members:
 
-
 :class:`~motor.motor_asyncio.AsyncIOMotorCommandCursor`
 =======================================================
 
--- a/doc/api-asyncio/cursors.rst
+++ b/doc/api-asyncio/cursors.rst
@@ -5,6 +5,7 @@
 
 .. autoclass:: AsyncIOMotorCursor
   :members:
+  :inherited-members:
 
 
 :class:`~motor.motor_asyncio.AsyncIOMotorCommandCursor`
@@ -14,3 +15,4 @@
 
 .. autoclass:: AsyncIOMotorCommandCursor
   :members:
+  :inherited-members:
--- a/doc/motor_extensions.py
+++ b/doc/motor_extensions.py
@@ -129,6 +129,13 @@ def get_motor_attr(motor_class, name, *defargs):
     attribute. While we're at it, store some info about each attribute
     in the global motor_info dict.
     """
+    try:
+        attr = safe_getattr(motor_class, name)
+    except AttributeError:
+        if name == '__annotations__':
+            return {}
+        raise
+
     attr = safe_getattr(motor_class, name)
 
     # Store some info for process_motor_nodes()
--- a/doc/motor_extensions.py
+++ b/doc/motor_extensions.py
@@ -129,14 +129,7 @@ def get_motor_attr(motor_class, name, *defargs):
     attribute. While we're at it, store some info about each attribute
     in the global motor_info dict.
     """
-    try:
-        attr = safe_getattr(motor_class, name)
-    except AttributeError:
-        if name == '__annotations__':
-            return {}
-        raise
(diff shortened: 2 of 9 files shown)


tox.ini
```diff
@@ -79,7 +79,7 @@ commands =
 setenv = PYTHONHASHSEED=0
 changedir = doc
 commands =
-    sphinx-build -q -E -b doctest . {envtmpdir}/doctest {posargs}
```
ShaneHarvey, line 82: what does removing `-q` do?

doc/api-asyncio/cursors.rst
```diff
@@ -5,7 +5,7 @@
 
 .. autoclass:: AsyncIOMotorCursor
   :members:
-
+  :inherited-members:
```
ShaneHarvey, line 8: Does this include anything that `members` doesn't? I don't see a difference in the rendered docs. FYI to get the docs to build with `inherited-members` add this workaround:
```
diff --git a/doc/motor_extensions.py b/doc/motor_extensions.py
index 16cbe11..75d1a13 100644
--- a/doc/motor_extensions.py
+++ b/doc/motor_extensions.py
@@ -129,7 +129,12 @@ def get_motor_attr(motor_class, name, *defargs):
     attribute. While we're at it, store some info about each attribute
     in the global motor_info dict.
     """
-    attr = safe_getattr(motor_class, name)
+    try:
+        attr = safe_getattr(motor_class, name)
+    except AttributeError:
+        if name == '__annotations__':
+            return {}
+        raise
```

tox.ini
prashantmital: `-q` instructs `sphinx-build` to not send any output to stdout (it stands for 'quiet'). Errors shows up on stderr. 
I don't see the point of this - the doctest output isn't great but its still useful to look at it. It might have helped us catch the fact that most of our doctests weren't running at all.

doc/api-asyncio/cursors.rst
prashantmital: Without `:inherited-members:`, I am not seeing docs for `fetch_next` and `next_object` in my local build.

Also, without `:inherited-members:` the doctests are not run:
```
Doctest summary
===============
  108 tests
    0 failures in tests
    0 failures in setup code
    0 failures in cleanup code
build succeeded.
```

With `:inherited-members:`:
```
Doctest summary
===============
  164 tests
    0 failures in tests
    0 failures in setup code
    0 failures in cleanup code
build succeeded.
```

Some tests are getting run multiple times (because e.g. of multiple classes inheriting from the same base class with a doctest) which is why there is such a large disparity in the number of test cases

prashantmital: Thanks for the workaround! I couldn't figure that out. I also have no clue why it even happens - it seems to be `tox`-specific because running `make html` or `make doctest` in `doc/` doesn't cause this failure.

prashantmital:
Failures are just synchro failing so you can review this @ShaneHarvey

doc/api-asyncio/cursors.rst
ShaneHarvey: FYI this is where the `__annotations__` error comes from: https://github.com/sphinx-doc/sphinx/blob/v3.2.0/sphinx/ext/autodoc/__init__.py#L624

doc/motor_extensions.py
```diff
@@ -129,6 +129,13 @@ def get_motor_attr(motor_class, name, *defargs):
...
     in the global motor_info dict.
     """
+    try:
+        attr = safe_getattr(motor_class, name)
+    except AttributeError:
+        if name == '__annotations__':
+            return {}
+        raise
+
     attr = safe_getattr(motor_class, name)
```
ShaneHarvey, line 139: Let's change this to `attr = safe_getattr(motor_class, name, *defargs)` instead of catching AttributeError.

prashantmital: Done.

PR #89, opened 2020-09-15
MOTOR-614 Use Python 3 syntax for super() calls

prashantmital:


motor/metaprogramming.py
```diff
@@ -278,7 +278,7 @@ def create_class_with_framework(cls, framework, module_name):
     cls_dict = dict(cls.__dict__)
     cls_dict.pop('__dict__', None)
 
-    new_class = type(str(motor_class_name), cls.__bases__, cls_dict)
+    new_class = type(str(motor_class_name), (cls,), cls_dict)
```
ShaneHarvey, line 281: I think we may be able to replace `cls_dict` with `{}` now.

ShaneHarvey (commented):
Can you run the entire test suite (synchro, docs, etc..) to be more sure this doesn't break anything?

ShaneHarvey (changes requested):
We should also test that this fixes MOTOR-614 by adding a test that successfully instantiates subclasses of client, database, and collection.

test/asyncio_tests/test_asyncio_basic.py
```diff
@@ -130,3 +130,19 @@ class C(ABC):
...
+        class DatabaseSubclass(motor_asyncio.AsyncIOMotorDatabase):
+            def __getitem__(self, name):
+                return CollectionSubclass(self, name)
+
+        class ClientSubclass(motor_asyncio.AsyncIOMotorClient):
+            def __getitem__(self, name):
+                return DatabaseSubclass(self, name)
+
+        cx = ClientSubclass(test.env.uri, **self.get_client_kwargs())
+        self.assertIsNotNone(await cx.testdb.testcoll.insert_one({}))
```
ShaneHarvey, line 148: Neat test! One bug is that `cx.testdb.testcoll` uses `__getattr__ ` not `__getitem__`. Can we assert that the database and collection are actually DatabaseSubclass and CollectionSubclass?

prashantmital: `__getattr__` for these classes actually calls `__getitem__` which is why I wrote it this way. But you are right, better to use the method directly to make the test less confusing. Updated and added type checks as well.

ShaneHarvey (approved):
LGTM, can you open a pymongo ticket for the failing changestream tests on latest?

PR #88, opened 2020-08-25
MOTOR-613: Add contextvars support

khasanovbi:
Add support to context vars.

This may be useful for logging, e.g. pass trace params like request_id to CommandLogger

PR #90, opened 2020-09-16
MOTOR-613 Test contextvars support

ShaneHarvey:
Added a test case inspired by the example in https://github.com/mongodb/motor/pull/88.

### DIFF
--- a/test/asyncio_tests/test_asyncio_client.py
+++ b/test/asyncio_tests/test_asyncio_client.py
@@ -19,9 +19,15 @@ import os
 import unittest
 from unittest import SkipTest
 
+try:
+    import contextvars
+except ImportError:
+    contextvars = False
+
+
 import pymongo
 from bson import CodecOptions
-from pymongo import ReadPreference, WriteConcern
+from pymongo import monitoring, ReadPreference, WriteConcern
 from pymongo.errors import ConnectionFailure, OperationFailure
 
 import motor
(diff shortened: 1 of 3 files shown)


test/asyncio_tests/test_asyncio_client.py
```diff
@@ -214,6 +220,44 @@ async def test_list_database_names(self):
...
+                self.vars = []
+
+            def save_contextvar_value(self):
+                self.vars.append(contextvar.get())
+
+            def started(self, event):
+                self.save_contextvar_value()
+
+            def succeeded(self, event):
+                self.save_contextvar_value()
```
prashantmital, line 239: Can we change (`set()`) the contextvar value here to 'succeeded' and check it after the await call has returned? Seems more compelling than just saving the same value over and over again.

test/asyncio_tests/test_asyncio_client.py
```diff
@@ -214,6 +220,44 @@ async def test_list_database_names(self):
...
+
+            def failed(self, event):
+                self.save_contextvar_value()
```
prashantmital, line 242: I think we can just pass here since we don't expect a failure in this particular test?

test/asyncio_tests/test_asyncio_client.py
```diff
@@ -214,6 +220,44 @@ async def test_list_database_names(self):
...
+        listener = Listener()
+        client = self.asyncio_client(event_listeners=[listener])
+        coll = client[self.db.name].test
+
+        await coll.insert_one({})
+        self.assertTrue(listener.vars)
+        for var in listener.vars:
+            self.assertEqual(var, 'default')
+
+        contextvar.set('ContextVar value')
```
prashantmital, line 253: What exactly does this test?

prashantmital (changes requested):

test/asyncio_tests/test_asyncio_client.py
ShaneHarvey: This is actually the use case we want to test. set() doesn't work like that because the executor gets a _copy_ of the context. The listener callbacks can make modifications to the context but those changes won't be reflected in the original context.

test/asyncio_tests/test_asyncio_client.py
ShaneHarvey: Done.

test/asyncio_tests/test_asyncio_client.py
ShaneHarvey: See below where we assert that the listener callbacks have access to the new value. This is the part that tests contextvars get forwarded to the background executors.

test/asyncio_tests/test_asyncio_client.py
```diff
@@ -214,6 +220,44 @@ async def test_list_database_names(self):
         self.assertIsInstance(names, list)
         self.assertIn(self.collection.database.name, names)
 
+    @unittest.skipIf(not contextvars, 'this test requires contextvars')
+    @asyncio_test
+    async def test_contextvars_support(self):
+        contextvar = contextvars.ContextVar('variable', default='default')
```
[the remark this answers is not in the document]
ShaneHarvey: Renamed to "var".

prashantmital (approved):

PR #96, opened 2021-01-13
MOTOR-633 watch methods must support MotorClientSession classes

ShaneHarvey:


test/asyncio_tests/test_asyncio_change_stream.py
```diff
@@ -260,3 +260,15 @@ async def test_database(self):
...
+
+    @asyncio_test
+    async def test_watch_with_session(self):
+        async with await self.cx.start_session() as session:
+            # Pass MotorSession.
+            async with self.collection.watch(session=session) as cs:
+                self.wait_and_insert(cs, 1)
+                _ = await cs.next()
+            # Pass PyMongo session directly.
+            async with self.collection.watch(session=session.delegate) as cs:
```
prashantmital, line 272: Why are we testing this? I don't think we want to advertise the fact that you can mix a PyMongo session with a Motor method call and still get the expected result. I recommend removing this test so that things don't start failing if `unwrap_kwargs_session` behavior changes.

prashantmital (changes requested):

test/asyncio_tests/test_asyncio_change_stream.py
ShaneHarvey: This ensures that we don't break backwards compatibility. Previously, the only workaround for this bug was to pass `session=session.delegate` directly. Breaking such code would just cause more pain.

prashantmital: Ok, got it. Still can you add a comment that we should stop testing this in 3.0?

prashantmital (approved):

test/asyncio_tests/test_asyncio_change_stream.py
ShaneHarvey: I created https://jira.mongodb.org/browse/MOTOR-642.

PR #97, opened 2021-01-13
MOTOR-625 Avoid exhausting worker thread pool when many tasks wait on ChangeStream.next

ShaneHarvey:
TODO: 

- [ ] Add a regression test that creates more ChangeStreams than MOTOR_MAX_WORKERS and calls next() on all of them. The test should timeout on 2.3 and pass on 2.3.1.


test/asyncio_tests/test_asyncio_change_stream.py
```diff
@@ -260,3 +263,24 @@ async def test_database(self):
             i += 1
             if i == 3:
                 break
+
+    @asyncio_test(timeout=10)
+    async def test_iterate_more_streams_than_workers(self):
```
ShaneHarvey, line 268: This test fails without the included fix. The failure looks like this:
```
ERROR [10.191s]: test_iterate_more_streams_than_workers (test.asyncio_tests.test_asyncio_change_stream.TestAsyncIOChangeStream)
  File "/Users/shane/git/motor/test/asyncio_tests/__init__.py", line 48, in __call__
    result = self.orig_method()
  File "/Users/shane/git/motor/test/asyncio_tests/__init__.py", line 213, in wrapped
    self.loop.run_until_complete(task)
  File "/Library/Frameworks/Python.framework/Versions/3.7/lib/python3.7/asyncio/base_events.py", line 587, in run_until_complete
    return future.result()
  File "/Library/Frameworks/Python.framework/Versions/3.7/lib/python3.7/asyncio/tasks.py", line 449, in wait_for
    raise futures.TimeoutError()
concurrent.futures._base.TimeoutError
```

motor/core.py
```diff
@@ -1664,19 +1663,24 @@ def _lazy_init(self):
...
+    @property
+    def alive(self):
+        """Does this cursor have the potential to return more data?
+
+        .. note:: Even if :attr:`alive` is ``True``, :meth:`next` can raise
+            :exc:`StopAsyncIteration` and :meth:`try_next` can return ``None``.
+
+        """
+        if not self.delegate:
+            # Not yet fully initialized, so we way return data.
```
prashantmital, line 1680: *may return data?

test/asyncio_tests/test_asyncio_change_stream.py
```diff
@@ -260,3 +263,24 @@ async def test_database(self):
...
+        # Create more tasks running ChangeStream.next than there are worker
+        # threads, and then ensure that other tasks can still run.
+        streams = [self.collection.watch() for _ in range(max_workers)]
+        tasks = [stream.next() for stream in streams]
```
prashantmital, line 272: How come this doesn't end up raising StopAsyncIteration() ? IIUC, `asyncio.gather` should propagate that error when it occurs...

motor/core.py
ShaneHarvey: Fixed.

test/asyncio_tests/test_asyncio_change_stream.py
ShaneHarvey: Notice the intentionally missing call to `await`. We're calling stream.next() to create a task but not actually awaiting it until the call to gather().

prashantmital: Right. But `gather` should `await` each of these and propagate the StopAsyncInteration exceptions raised therein because we are not setting `return_exceptions=True`. As per docs of `gather`: 

> If return_exceptions is False (default), the first raised exception is immediately propagated to the task that awaits on gather(). Other awaitables in the aws sequence won’t be cancelled and will continue to run.

test/asyncio_tests/test_asyncio_change_stream.py
```diff
@@ -260,3 +263,24 @@ async def test_database(self):
...
+        try:
+            async def find_insert():
+                # Wait for all change streams to be created
+                while not all(stream.delegate for stream in streams):
+                    await asyncio.sleep(.1)
+                await self.collection.find_one()
+                await self.collection.insert_one({})
+            tasks.extend([find_insert() for _ in range(10)])
+            await asyncio.gather(*tasks)
```
prashantmital, line 281: I would expect this command to raise an exception that will cause the test fail. Why doesn't it?

test/asyncio_tests/test_asyncio_change_stream.py
ShaneHarvey: The next() call returns a document because we insert documents to the same collection.

prashantmital (approved):
LGTM

ShaneHarvey:
Rebased to fix merge conflicts.

PR #105, opened 2021-04-01
MOTOR-607: Improve how we raise DeprecationWarnings in Motor

guanlinzhou:
Surface `DeprecationWarnings` to user without using python flags
```
(base) william.zhou@M-C02F15QFMD6M ~ % cat sample.py
import asyncio
 
async def main():
    client = AsyncIOMotorClient()
    print(await client.admin.command('ping'))
    print(client.nodes)
    await client.fsync()
 
asyncio.run(main())
(base) william.zhou@M-C02F15QFMD6M ~ % python sample.py    
{'ok': 1.0}
frozenset({('localhost', 27017)})
/Users/william.zhou/opt/anaconda3/lib/python3.8/concurrent/futures/thread.py:57: DeprecationWarning: fsync is deprecated. Use client.admin.command('fsync') instead.
(base) william.zhou@M-C02F15QFMD6M ~ % python -Wd sample.py
{'ok': 1.0}
frozenset({('localhost', 27017)})
/Users/william.zhou/opt/anaconda3/lib/python3.8/concurrent/futures/thread.py:57: DeprecationWarning: fsync is deprecated. Use client.admin.command('fsync') instead.
```

### DIFF
--- a/motor/core.py
+++ b/motor/core.py
@@ -56,8 +56,9 @@ except ImportError:
     ssl = None
     HAS_SSL = False
 
-# Surface deprecation warnings to users by default
-warnings.filterwarnings("default", category=DeprecationWarning)
+# Surface warnings to users by default
+if not sys.warnoptions:
+    warnings.simplefilter("default")
 
 # From the Convenient API for Transactions spec, with_transaction must
 # halt retries after 120 seconds.
--- a/motor/core.py
+++ b/motor/core.py
@@ -56,6 +56,9 @@ except ImportError:
     ssl = None
     HAS_SSL = False
 
+# Surface deprecation warnings to users by default
+warnings.filterwarnings("default", category=DeprecationWarning)
+
 # From the Convenient API for Transactions spec, with_transaction must
 # halt retries after 120 seconds.
 # This limit is non-configurable and was chosen to be twice the 60 second


motor/core.py
```diff
@@ -56,6 +56,9 @@
     ssl = None
     HAS_SSL = False
 
+# Surface deprecation warnings to users by default
+warnings.filterwarnings("default", category=DeprecationWarning)
```
ShaneHarvey, line 60: What are the implications of this? IIUC we are changing this option for the entire Python process, not just Motor. For example what happens when the user runs with `python -Werror ..`? Does this override their "error" setting?

What if we manually add deprecation warnings to motor itself before attempting the operation on the background thread?

guanlinzhou: > What are the implications of this? IIUC we are changing this option for the entire Python process, not just Motor. For example what happens when the user runs with `python -Werror ..`? Does this override their "error" setting?
> 
DeprecationWarnings will still show as warnings in that case I think so it does do overriding
> What if we manually add deprecation warnings to motor itself before attempting the operation on the background thread?
What exactly does this entail?

ShaneHarvey: Motor essentially forwards method calls to a background thread which executes them via Pymongo on a background thread. You can read more about Motor's design here: https://motor.readthedocs.io/en/stable/developer-guide.html

I'm suggesting that _before_ sending a deprecated method to a background thread, we explicitly call `warnings.warn('...', DeprecationWarning, stacklevel=...)` in motor. This isn't ideal since we need to maintain these duplicate warnings in motor but it might be the best we can do. 

An alternative might be to somehow forward the warning from the background thread but I'm not sure it that is feasible.

ShaneHarvey: @prashantmital can you chime in. Is this required for 2.4? Maybe we should punt this issue to the next release?

guanlinzhou: ok I'll leave this open for now and work on something else

prashantmital: I think `sys.warnoptions` is what we need here. 

```
if not sys.warnoptions:
    warnings.simplefilter("default")
```

(see https://docs.python.org/3/library/warnings.html?highlight=warnings#overriding-the-default-filter for another example of using `sys.warnoptions`)


I am also not sure why we are restricting this to `DeprecationWarning`? Seems like this would be an issue that plagues any warnings raised by a background thread. Hence, I suggested the above solution that would do this for _all_ warnings but maybe I am missing something here.

guanlinzhou: Prashant's fix seems to fix the cases where users specify a flag (such as `-Werror`) and allow them to display as the user wants them to

ShaneHarvey: It seems odd to me that Motor, a python library (not an application), would change the default warning filter for the entire application. [MOTOR-607](https://jira.mongodb.org/browse/MOTOR-607) says 
> Since these synchronous calls are run using a thread pool executor, the application must be run using -Wdefault in order to make the warning user-visible.

Is this behavior explained anywhere? I can't fin anything about threads not showing warnings as expected.

ShaneHarvey: We definitely cannot call `warnings.simplefilter("default")`. Wouldn't it break applications that do this:
```
warnings.simplefilter("error")
import motor
# Calling deprecated method should error but does not...
```

guanlinzhou: I believe `DeprecationWarnings` are ignored by Python default and not anything motor is doing 
image

guanlinzhou: I'll run a quick test on this

> We definitely cannot call `warnings.simplefilter("default")`. Wouldn't it break applications that do this?:
> 
> ```
> warnings.simplefilter("error")
> import motor
> # Calling deprecated method should error but does not...
> ```

guanlinzhou: @ShaneHarvey you're right, it does break those applications

prashantmital: > Is this behavior explained anywhere? I can't fin anything about threads not showing warnings as expected.

I think @guanlinzhou is right - `DeprecationWarning`s seem to be suppressed unless they are raised in `__main__`. So it appears that this is expected behavior.

prashantmital (changes requested):
This seems to be expected behavior so let's close it.

PR #107, opened 2021-04-06
MOTOR-701: Require MongoDB 4.2+ to run CSFLE tests

guanlinzhou:
Older versions of MDB do not support CSFLE, so require 4.2 version to run tests


prashantmital (approved):
The change LGTM but can you:
- Update the PR name to "Require MongoDB 4.2+ to run CSFLE tests"
- Update the MOTOR-701 ticket name to the same^

ShaneHarvey (approved):

PR #108, opened 2021-04-06
MOTOR-702: Motor tests that run agains PyMongo master should use v3.12 branch

guanlinzhou:
Change `tox.ini` to use v3.12 branch instead of master branch for testing against PyMongo

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -23,8 +23,8 @@ envlist =
     # asyncio without Tornado.
     asyncio-{pypy35,pypy36,py35,py36,py37,py38,py39},
 
-    # Test PyMongo HEAD, not the latest release.
-    py3-pymongo-master,
+    # Test PyMongo 3.12 because 4.x breaks some APIs
+    py3-pymongo-v3.12,
 
     # Apply PyMongo's test suite to Motor via Synchro.
     synchro37
@@ -61,8 +61,8 @@ deps =
     sphinx: aiohttp
     sphinx: git+https://github.com/tornadoweb/tornado.git
 
-    py3-pymongo-master: git+https://github.com/mongodb/mongo-python-driver.git
-    py3-pymongo-master: tornado>=5,<6
+    py3-pymongo-v3.12: git+https://github.com/mongodb/mongo-python-driver.git@v3.12
+    py3-pymongo-v3.12: tornado>=5,<6
 
     synchro37: tornado>=6,<7
     synchro37: nose
@@ -91,5 +91,5 @@ whitelist_externals =
 setenv =
     PYTHONPATH = {envtmpdir}/mongo-python-driver
 commands =
-    git clone --depth 1 https://github.com/mongodb/mongo-python-driver.git {envtmpdir}/mongo-python-driver
+    git clone --depth 1 https://github.com/mongodb/mongo-python-driver.git@v3.12 {envtmpdir}/mongo-python-driver
     python3 -m synchro.synchrotest --with-xunit --xunit-file=xunit-synchro-results -v -w {envtmpdir}/mongo-python-driver {posargs}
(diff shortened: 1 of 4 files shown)


ShaneHarvey (commented):
Do we need to make some evergreen config changes to match the tox environment name changes?

tox.ini
```diff
@@ -91,5 +91,5 @@ whitelist_externals =
 setenv =
     PYTHONPATH = {envtmpdir}/mongo-python-driver
 commands =
-    git clone --depth 1 https://github.com/mongodb/mongo-python-driver.git {envtmpdir}/mongo-python-driver
+    git clone --depth 1 https://github.com/mongodb/mongo-python-driver.git@v3.12 {envtmpdir}/mongo-python-driver
```
prashantmital, line 94: I think this is `pip` syntax and might not work in a `git clone` command. For this you likely need to do
```
git clone --branch v3.12 https://github.com/mongodb/mongo-python-driver.git
```

prashantmital (changes requested):

tox.ini
guanlinzhou: ahh ok updated!

guanlinzhou:
> Do we need to make some evergreen config changes to match the tox environment name changes?

Changed some evergreen config to test!

prashantmital (changes requested):
Can you schedule the `synchro37` task on the Evergreen build for this commit? 

Given what we discovered in MOTOR-703, this PR only _really_ affects the pymongo version that the `synchro37` tests  are using. The `py3-pymongo-master`/`py3-pymongo-v3.12` tests are simply using whatever PyMongo version is specified under `install_requires` in `setup.py`.

tox.ini
```diff
@@ -23,7 +23,11 @@ envlist =
     # asyncio without Tornado.
     asyncio-{pypy35,pypy36,py35,py36,py37,py38,py39},
 
+<<<<<<< HEAD
```
prashantmital: Looks like you didn't merge this file correctly - can you fix it? It should just read:
```
# Test PyMongo 3.12 because 4.x breaks some APIs
py3-pymongo-v3.12,
```

tox.ini
```diff
@@ -61,6 +65,7 @@ deps =
     sphinx: aiohttp
     sphinx: git+https://github.com/tornadoweb/tornado.git
 
+    py3-pymongo-v3.12: git+https://github.com/mongodb/mongo-python-driver.git@v3.12
```
prashantmital, line 64: You can remove this - we now install pymongo for this test in the `commands` section.

tox.ini
```diff
@@ -97,5 +102,7 @@ whitelist_externals =
 setenv =
     PYTHONPATH = {envtmpdir}/mongo-python-driver
 commands =
-    git clone --depth 1 https://github.com/mongodb/mongo-python-driver.git {envtmpdir}/mongo-python-driver
+    pip install git+https://github.com/mongodb/mongo-python-driver.git@v3.12#egg=pymongo[encryption]
```
prashantmital, line 101: This is incorrect. I don't think synchro can work like this... Your previous changes were correct for this... Something like:
```
git clone --depth 1 --branch v3.12 https://github.com/mongodb/mongo-python-driver.git {envtmpdir}/mongo-python-driver
```

prashantmital (changes requested):

tox.ini
guanlinzhou: removed this already

tox.ini
guanlinzhou: ok, I'll retest but the a lot of errors from synchro were poppiing up

prashantmital (approved):

ShaneHarvey (approved):

PR #109, opened 2021-04-06
MOTOR-703 Ensure we are testing against latest pymongo v3.12 commit

prashantmital:
We must override the PyMongo dependency as part of [tox commands](https://tox.readthedocs.io/en/latest/index.html#system-overview) to ensure the version specified in `install_requires` doesn't prevail. The `#egg=pymongo[encryption]` is required to pull in the `encryption` extra with the `pip git+` syntax.

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -24,7 +24,7 @@ envlist =
     asyncio-{pypy35,pypy36,py35,py36,py37,py38,py39},
 
     # Test PyMongo HEAD, not the latest release.
-    py3-pymongo-master,
+    py3-pymongo-v3.12,
 
     # Apply PyMongo's test suite to Motor via Synchro.
     synchro37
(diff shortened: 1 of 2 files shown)


tox.ini
```diff
@@ -85,6 +84,13 @@ changedir = doc
 commands =
     sphinx-build -q -E -b doctest . {envtmpdir}/doctest {posargs}
 
+[testenv:py3-pymongo-v3.12]
+commands =
+    pip install git+https://github.com/mongodb/mongo-python-driver.git#egg=pymongo[encryption]
```
ShaneHarvey, line 89: Doesn't this install pymongo@master? Should we use `git clone` or `pip install https://github.com/mongodb/mongo-python-driver/archive/v3.12.tar.gz` instead?

ShaneHarvey (approved):
LGTM, can you open a motor ticket for the one test failure?:
```
   File "/data/mci/b42ec2f6c7753616ea9aa20d003f18ab/src/.tox/py3-pymongo-v3.12/lib/python3.6/site-packages/tornado/testing.py", line 125, in __call__
     result = self.orig_method(*args, **kwargs)
   File "/data/mci/b42ec2f6c7753616ea9aa20d003f18ab/src/test/tornado_tests/test_motor_core.py", line 90, in test_client_attrs
     attrs(self.cx) - motor_client_only)
 AssertionError: Items in the first set but not the second:
 'topology_description'
 ```

PR #116, opened 2021-06-07
MOTOR-747 Update test certs from drivers-evergreen-tools

ShaneHarvey:
See https://jira.mongodb.org/browse/MOTOR-747


behackett:
Are these the certs from PyMongo?

ShaneHarvey:
They are copied from drivers-evergreen-tools. These are the certs that we actually test with: https://github.com/mongodb/motor/blob/fad68810f918b48da41bfce0e4e4c363ac33a7c9/.evergreen/run-tox.sh#L23-L24

prashantmital (approved):

PR #119, opened 2021-07-15
MOTOR-769  MOTOR-770 Fix test_fetch_next_exception and test_exhaust

ShaneHarvey:


ShaneHarvey:
Both of these test failures were caused because the tests muck around with internal pymongo state around cursor closing and cursor del. The way we handle those changed subtly in PyMongo 3.12 so the tests started failing. There's no pymongo bug here.

ShaneHarvey:
The Windows failures are expected.

test/asyncio_tests/test_asyncio_cursor.py
```diff
@@ -467,16 +467,24 @@ async def test_exhaust(self):
         sock = one(socks)
         cur = client[self.db.name].test.find(
             cursor_type=CursorType.EXHAUST).batch_size(1)
-        has_next = await cur.fetch_next
-        self.assertTrue(has_next)
+        await cur.fetch_next
+        self.assertTrue(cur.next_object())
```
prashantmital, line 471: Can't we write this test in a way that it doesn't use these deprecated methods? We really should be testing this with `async for`

test/asyncio_tests/test_asyncio_grid_file.py
```diff
@@ -122,9 +122,6 @@ async def test_readchunk(self):
 
     @asyncio_test
     async def test_gridout_open_exc_info(self):
-        if sys.version_info < (3, ):
```
prashantmital, line 125: Nice 👍🏽

prashantmital (commented):
Do we have a ticket already to eliminate duplication of test code between asyncio and tornado test modules? It is entirely unnecessary now that we only support Python 3 and Tornado has deprecated its own coroutine syntax in favor of asyncio so the APIs for concrete API classes are now identical...

test/asyncio_tests/test_asyncio_cursor.py
ShaneHarvey: Yes we could but I think that's out of scope for this ticket.

prashantmital: It is? I was just suggesting not adding any new test code that uses deprecated methods, but OK.

prashantmital (approved):

ShaneHarvey:
> Do we have a ticket already to eliminate duplication of test code between asyncio and tornado test modules? It is entirely unnecessary now that we only support Python 3 and Tornado has deprecated its own coroutine syntax in favor of asyncio so the APIs for concrete API classes are now identical...

I opened https://jira.mongodb.org/browse/MOTOR-772

test/asyncio_tests/test_asyncio_cursor.py
ShaneHarvey: This test is already using fetch_next/next_object. Using async-for would be out of place and rewriting the whole test to use async-for is a much larger change, something that should happen in MOTOR-605.

PR #121, opened 2021-07-15
MOTOR-771 Use Sphinx 4 on motor.readthedocs.io

ShaneHarvey:
See https://docs.readthedocs.io/en/stable/guides/reproducible-builds.html#using-a-configuration-file

### DIFF
--- a/doc/docs-requirements.txt
+++ b/doc/docs-requirements.txt
@@ -1,2 +1,5 @@
 tornado
 aiohttp
+Sphinx~=4.1
+sphinx_rtd_theme~=0.5
+readthedocs-sphinx-search~=0.1
--- a/.readthedocs.yaml
+++ b/.readthedocs.yaml
@@ -1,0 +1,16 @@
+# .readthedocs.yaml
+# Read the Docs configuration file
+# See https://docs.readthedocs.io/en/stable/config-file/v2.html for details
+
+# Required
+version: 2
+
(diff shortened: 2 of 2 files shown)


doc/docs-requirements.txt
```diff
@@ -1,2 +1,5 @@
 tornado
 aiohttp
+Sphinx~=4.1
```
prashantmital, line 3: What is `~=`?

doc/docs-requirements.txt
```diff
@@ -1,2 +1,5 @@
...
+sphinx_rtd_theme~=0.5
```
prashantmital, line 4: Are these not implicitly specified by installing `Sphinx`? Why do they need to be explicitly specified now?

ShaneHarvey: They are not. These are readthedocs specific dependencies. The docs recommend pinning them: https://docs.readthedocs.io/en/stable/guides/reproducible-builds.html#pinning-dependencies

doc/docs-requirements.txt
ShaneHarvey:
```
Mopidy-Dirble ~= 1.1        # Compatible release. Same as >= 1.1, == 1.*
```
See https://pip.pypa.io/en/stable/cli/pip_install/#example-requirements-file

prashantmital (approved):

PR #122, opened 2021-07-15
MOTOR-779 Test against MongoDB 5.0

ShaneHarvey:


behackett (approved):

ShaneHarvey:
Turns out 5.0 doesn't support Ubuntu 16 so we need to migrate to something else. Going to choose amazon1-2018-test like we did in pymongo.

PR #124, opened 2021-07-16
MOTOR-730 Stop testing problematic .js content type

ShaneHarvey:
See https://bugs.python.org/issue43975.

### DIFF
--- a/test/asyncio_tests/test_aiohttp_gridfs.py
+++ b/test/asyncio_tests/test_aiohttp_gridfs.py
@@ -222,7 +222,6 @@ class AIOHTTPGridFSHandlerTest(AIOHTTPGridFSHandlerTestBase):
             ('bar', 'octet-stream'),
             ('bar.png', 'png'),
             ('ht.html', 'html'),
-            ('jscr.js', 'javascript'),
         ]:
             # 'fs' is PyMongo's blocking GridFS
             _id = self.fs.put(b'', filename=filename)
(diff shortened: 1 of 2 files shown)


test/asyncio_tests/test_aiohttp_gridfs.py
```diff
@@ -222,7 +222,6 @@ async def test_content_type(self):
             ('bar', 'octet-stream'),
             ('bar.png', 'png'),
             ('ht.html', 'html'),
-            ('jscr.js', 'javascript'),
```
ShaneHarvey, line 225: I don't think we loose anything by removing javascript here. We already test various other types.

test/tornado_tests/test_motor_web.py
```diff
@@ -169,17 +169,16 @@ def test_content_type(self):
...
             for method in 'GET', 'HEAD':
                 response = self.fetch('/' + filename, method=method)
                 self.assertEqual(200, response.code)
                 # mimetypes are platform-defined, be fuzzy
-                self.assertTrue(
-                    response.headers['Content-Type'].lower().endswith(
-                        expected_type))
+                self.assertIn(
+                    expected_type,
+                    response.headers['Content-Type'].lower())
```
ShaneHarvey, line 181: This is a copy/paste from test_aiohttp_gridfs.py to give a better error message when the assertion fails.

prashantmital (approved):

PR #126, opened 2021-07-22
MOTOR-735 Versioned API connection examples for ecosystem docs

prashantmital:


setup.py
```diff
@@ -30,7 +30,7 @@
 with open("README.rst") as readme:
     long_description = readme.read()
 
-pymongo_ver = ">=3.11,<4"
+pymongo_ver = ">=3.12,<4"
```
prashantmital, line 33: Needed to bump this here since 3.11 doesn't have `pymongo.server_api`. One less ticket!

test/asyncio_tests/test_examples.py
```diff
@@ -1107,3 +1108,31 @@ async def test_causal_consistency(self):
             async for item in items.find({'end': None}, session=s2):
                 print(item)
         # End Causal Consistency Example 2
+
+    @env.require_version_min(4, 7)
+    @asyncio_test
+    def test_versioned_api(self):
+        # Versioned API examples
+        AsyncIOMotorClient = lambda _, server_api: self.asyncio_client(
+            server_api=server_api, connect=False)
```
prashantmital, line 1117: I wrote this to mimic the PyMongo PR - https://github.com/mongodb/mongo-python-driver/commit/048f54ddde999459d27ff6d5a6fc5db0de28a490 
But I have no idea why we set `connect=False`? Can we remove it?

setup.py
ShaneHarvey: Great. Can you link MOTOR-699 in the commit message?

test/asyncio_tests/test_examples.py
ShaneHarvey: It's to reduce the overhead of the test. Since the test doesn't need to actually use the MongoClient there's no point in starting all the monitor threads. Let's keep it and add a comment.

setup.py
prashantmital: Will do! Thanks for pulling up the ticket.

test/asyncio_tests/test_examples.py
prashantmital: Ok

ShaneHarvey (approved):

PR #132, opened 2021-07-22
MOTOR-787 Remove support for Python 3.5

prashantmital:
PyMongo 3.12 dropped Python 3.5 so it makes sense to do the same for Motor 2.5. Rest of the deprecated API changes to come in a follow on PR.


ShaneHarvey:
PyMongo 3.12 did not drop support for Python 3.5.

prashantmital:
Sorry, I meant deprecated... Ah right... So is this entirely unnecessary?

ShaneHarvey:
Yes, I don't think we should drop 3.5 in this release. Maybe the next one.

PR #134, opened 2021-08-10
MOTOR-805 Fix that the number of query results is inconsistent with the length argument when calling to_list().

ephe-xiao:
Fix that the number of query results is inconsistent with the length argument when calling to_list().


motor/core.py
```diff
@@ -1434,10 +1434,11 @@ def _to_list(self, length, the_list, future, get_more_result):
             if reached_length or not self.alive:
                 future.set_result(the_list)
             else:
+                next_length = length - len(the_list)
```
ShaneHarvey, line 1437: This isn't quite right since it will subtract *too much* from the length on the 3rd, 4th, etc run. Instead of updating the length here I think it would be cleaner to change the line above like this:
```python
            if length is None:
                n = result
            else:
-                n = min(length, result)
+                n = min(length-len(the_list), result)
```

ShaneHarvey (changes requested):
Thanks for finding this! I'm surprised we don't already have test coverage for this case. Do you mind adding a test case? You can use "test_to_list_with_length" as a template here: https://github.com/mongodb/motor/blob/bdb970378aa7012ef632300fb7057059d0e7b6af/test/asyncio_tests/test_asyncio_cursor.py#L182

ShaneHarvey:
In order to reproduce the issue I ended up write this test:
```python
    @asyncio_test
    async def test_to_list_multiple_getMores(self):
        await self.make_test_data()
        coll = self.collection
        cursor = coll.find(batch_size=5).sort('_id')

        def expected(start, stop):
            return [{'_id': i} for i in range(start, stop)]

        # 2 batches (find+getMore):
        self.assertEqual(expected(0, 10), (await cursor.to_list(10)))
        # 5 batches, stop in the middle of a batch:
        self.assertEqual(expected(10, 33), (await cursor.to_list(23)))
        # 33 batches:
        self.assertEqual(expected(33, 200), (await cursor.to_list(167)))
        # Nothing left.
        self.assertEqual([], (await cursor.to_list(100)))
        await cursor.close()
```

motor/core.py
ephe-xiao: Ah, yes, I overlooked this way of batch modification

ephe-xiao:
> In order to reproduce the issue I ended up writing this test:
> 
> ```python
>     @asyncio_test
>     async def test_to_list_multiple_getMores(self):
>         await self.make_test_data()
>         coll = self.collection
>         cursor = coll.find(batch_size=5).sort('_id')
> 
>         def expected(start, stop):
>             return [{'_id': i} for i in range(start, stop)]
> 
>         # 2 batches (find+getMore):
>         self.assertEqual(expected(0, 10), (await cursor.to_list(10)))
>         # 5 batches, stop in the middle of a batch:
>         self.assertEqual(expected(10, 33), (await cursor.to_list(23)))
>         # 33 batches:
>         self.assertEqual(expected(33, 200), (await cursor.to_list(167)))
>         # Nothing left.
>         self.assertEqual([], (await cursor.to_list(100)))
>         await cursor.close()
> ```
> 
> It can go alongside the existing test_to_list_with_length test here:
> 
> https://github.com/mongodb/motor/blob/bdb970378aa7012ef632300fb7057059d0e7b6af/test/asyncio_tests/test_asyncio_cursor.py#L182

ShaneHarvey:
This is fixed in https://github.com/mongodb/motor/pull/135

PR #141, opened 2021-12-21
MOTOR-643 Support PyMongo 4.0

blink1073:


motor/aiohttp/__init__.py
```diff
@@ -190,7 +191,18 @@ async def __call__(self, request):
             raise aiohttp.web.HTTPNotFound(text=request.path)
 
         resp = aiohttp.web.StreamResponse()
-        self._set_standard_headers(request.path, resp, gridout)
+
+         # Calculate the md5 for the GridFS file.
+        md5 = hashlib.md5()
```
behackett, line 196: We shouldn't generate md5 sums. See https://jira.mongodb.org/browse/MOTOR-192, and there should be (but it doesn't look like there is) a motor equivalent to https://jira.mongodb.org/browse/PYTHON-2998.

blink1073: I've removed the `disable_md5` option and `md5` properties everywhere, in this case it is used for the [`Etag`](https://httpwg.org/specs/rfc7232.html#header.etag) `http` header, which is for the web server's use primarily.   The [spec](https://httpwg.org/specs/rfc7232.html#header.etag) doesn't dictate what algorithm to use, but `tornado` uses `sha1` and `CRC32C` is another good candidate (recommended by [Google Cloud](https://cloud.google.com/storage/docs/hashes-etags)).

blink1073: Actually, if the motivation is strong FIPS [compliance](https://www.ibm.com/docs/sr/itcam-app-mgr/7.2.1?topic=aic-complying-fips-requirements), we should use `sha256`, I'll update to that.

motor/aiohttp/__init__.py
```diff
@@ -190,7 +191,19 @@ async def __call__(self, request):
...
+        # Calculate the sha256 checksum for the GridFS file
+        # for a FIPS-compliant Etag HTTP header.
+        sha = hashlib.sha256()
+        while True:
+            chunk = await gridout.readchunk()
+            if not chunk:
+                break
+            sha.update(chunk)
+        gridout.seek(0)
+        sha = sha.hexdigest()
```
ShaneHarvey, line 204: We can't merge this change since it makes this method read the entire file twice. It also still reads the entire file even just to return a HTTP 304 Not Modified code.  https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/ETag says:

> The ETag (or entity tag) HTTP response header is an identifier for a specific version of a resource. It lets caches be more efficient and save bandwidth, as a web server does not need to resend a full response if the content was not changed. Additionally, etags help to prevent simultaneous updates of a resource from overwriting each other ("mid-air collisions").
> If the resource at a given URL changes, a new Etag value must be generated. A comparison of them can determine whether two representations of a resource are the same. Etags are therefore similar to fingerprints, and might also be used for tracking purposes by some servers. They might also be set to persist indefinitely by a tracking server.

I propose to hash the "Files collection document" or even a subset of the fields instead. GridFS files are supposed to be immutable so we don't have to worry about the contents changing. The only issue is if a user deletes and re-uploads a gridfs file with the same _id. In that case other fields in the files document (like uploadDate and length) would most likely change as well:
```
{
  "_id" : <TFileId>,
  "length" : <Int64>,
  "chunkSize" : <Int32>,
  "uploadDate" : <BSON datetime, ms since Unix epoch in UTC>,
  "md5" : <hex string>,
  "filename" : <string>,
  "contentType" : <string>,
  "aliases" : <string array>,
  "metadata" : <Document>
}
```
https://github.com/mongodb/specifications/blob/master/source/gridfs/gridfs-spec.rst

You can hash these fields on the GridOut:
```
    _id = _grid_out_property("_id", "The ``'_id'`` value for this file.")
    filename = _grid_out_property("filename", "Name of this file.")
    name = _grid_out_property("filename", "Alias for `filename`.")
    content_type = _grid_out_property("contentType", "Mime-type for this file.")
    length = _grid_out_property("length", "Length (in bytes) of this file.")
    chunk_size = _grid_out_property("chunkSize", "Chunk size for this file.")
    upload_date = _grid_out_property("uploadDate",
                                     "Date that this file was first uploaded.")
    aliases = _grid_out_property("aliases", "List of aliases for this file.")
    metadata = _grid_out_property("metadata", "Metadata attached to this file.")
    md5 = _grid_out_property("md5", "MD5 of the contents of this file "
                             "if an md5 sum was created.")
```

It might even be enough to hash only _id+length+upload_date. What do you think?

motor/motor_gridfs.py
```diff
@@ -277,10 +275,6 @@ def __init__(self, root_collection, delegate=None, session=None,
           - `session` (optional): a
             :class:`~pymongo.client_session.ClientSession` to use for all
             commands
-          - `disable_md5` (optional): When True, an MD5 checksum will not be
```
ShaneHarvey, line 280: Here and elsewhere please add a `.. versionchanged:: 3.0` to document that disable_md5 was removed.

test/asyncio_tests/test_asyncio_client.py
```diff
@@ -67,6 +68,8 @@ async def test_close(self):
     async def test_unix_socket(self):
         if env.mongod_started_with_ssl:
             raise SkipTest("Server started with SSL")
+        if os.name == "nt":
+            raise SkipTest("Unix socket not available on Windows")
```
ShaneHarvey, line 72: Why this change? The windows tests are passing in EVG: https://evergreen.mongodb.com/waterfall/motor

test/asyncio_tests/test_asyncio_client.py
```diff
@@ -67,6 +68,8 @@ async def test_close(self):
     async def test_unix_socket(self):
         if env.mongod_started_with_ssl:
             raise SkipTest("Server started with SSL")
+        if os.name == "nt":
```
ShaneHarvey, line 71: Please use `sys.platform == "win32"`. Hoping we can keep using one way to detect windows across all our repros.

test/tornado_tests/test_motor_core.py
```diff
@@ -169,10 +141,14 @@ async def test_gridout_attrs(self):
             'stream_to_handler'
         ]).union(motor_only)
 
+        gridin_only = set([
+            'md5', 'readlines', 'truncate', 'flush', 'fileno', 'closed', 'writelines',
+            'isatty', 'writable'])
```
ShaneHarvey, line 146: Can you open a new ticket to add support for these to motor? Or does it not make sense to implement IOBase in async world? Is there a std lib asyncio version of IOBase?

tox.ini
```diff
@@ -24,7 +24,7 @@ envlist =
     asyncio-{pypy35,pypy36,py35,py36,py37,py38,py39,py310},
 
     # Test PyMongo 3.12 because 4.x breaks some APIs
```
ShaneHarvey, line 26: We can update this comment now.

ShaneHarvey (changes requested):

motor/aiohttp/__init__.py
ShaneHarvey: Nice catch! I responded to this thread here: https://github.com/mongodb/motor/pull/141#discussion_r774145847

test/tornado_tests/test_motor_core.py
blink1073: These methods are unsupported in `pymongo` as [well](https://github.com/mongodb/mongo-python-driver/blob/c760f900f2e4109a247c2ffc8ad3549362007772/gridfs/grid_file.py#L680)

motor/aiohttp/__init__.py
blink1073: Sounds reasonable to me

motor/motor_gridfs.py
blink1073: Will do

test/asyncio_tests/test_asyncio_client.py
blink1073: ACK

test/asyncio_tests/test_asyncio_client.py
blink1073: Removing the skip

tox.ini
blink1073: Done

motor/motor_gridfs.py
blink1073: Done

ShaneHarvey (commented):
Just one minor issue otherwise LGTM!

ShaneHarvey (approved):

PR #154, opened 2022-04-12
MOTOR-922 Update docs.mongodb.com links in source, API & Reference documentation

blink1073:


doc/api-asyncio/asyncio_gridfs.rst
```diff
@@ -3,7 +3,7 @@ asyncio GridFS Classes
 
 .. currentmodule:: motor.motor_asyncio
 
-Store blobs of data in `GridFS <http://dochub.mongodb.org/core/gridfs>`_.
```
ShaneHarvey, line 6: We use dochub link on purpose because they are stable links that will redirect to the correct docs page. Can you ask the docs team if we should keep using dochub.mongodb.org or if there's a new alternative?

ShaneHarvey (approved):

PR #155, opened 2022-04-13
MOTOR-902 Since Python 3.10 asyncio.get_event_loop() is deprecated

blink1073:


blink1073:
Updated the tutorial to use `client.get_io_loop()` because there is no running loop yet until we run some async code.

motor/core.py
```diff
@@ -1042,22 +1048,13 @@ def main():
...
                   pass
-              finally:
-                  if change_stream is not None:
-                      change_stream.close()
-
-                  # Prevent "Task was destroyed but it is pending!"
-                  loop.run_until_complete(task)
+          finally:
+              if change_stream is not None:
+                  change_stream.close()
```
ShaneHarvey, line 1057: change_stream.close() is an async method so calling it here leaks a coroutine. It's fine for this PR since it's a preexisting issue but could you file a new issue to improve this code example? Really it should work via task cancellation, not through KeyboardInterrupt.

pyproject.toml
```diff
@@ -0,0 +1,3 @@
+[build-system]
+requires = ["setuptools>=59.0"]
```
ShaneHarvey, line 2: Why do we need setuptools>=59.0? And should we add "wheel"?

tox.ini
```diff
@@ -70,6 +70,8 @@ deps =
     synchro37: tornado>=6,<7
     synchro37: nose
 
+setenv =
+    PYTHONWARNINGS="error,ignore:The distutils package is deprecated:DeprecationWarning"
```
ShaneHarvey, line 74: Do we still need to ignore this now that we use setuptools on >=3.10?

motor/frameworks/asyncio/__init__.py
```diff
@@ -37,7 +37,11 @@
 
 
 def get_event_loop():
-    return asyncio.get_event_loop()
+    try:
+        return asyncio.get_running_loop()
+    except RuntimeError:
+        # Workaround for bugs.python.org/issue39529.
+        return asyncio.get_event_loop_policy().get_event_loop()
```
ShaneHarvey, line 44: I was expecting `asyncio.set_event_loop(asyncio.new_event_loop())`. Doesn't `AbstractEventLoopPolicy.get_event_loop()` have the same behavior as asyncio.get_event_loop()?

Did the maintainers just forget to deprecate AbstractEventLoopPolicy.get_event_loop?

ShaneHarvey (changes requested):

motor/core.py
```diff
@@ -146,7 +147,12 @@ def __init__(self, *args, **kwargs):
 
         delegate = self.__delegate_class__(*args, **kwargs)
         super().__init__(delegate)
-        self.io_loop = io_loop
+
+    @property
+    def io_loop(self):
+        if self._io_loop is None:
+            self._io_loop = self._framework.get_event_loop()
```
ShaneHarvey, line 154: we might need to make a similar change in AgnosticClientEncryption.__init__.

tox.ini
blink1073: The warning is still raised by [pip itself](https://github.com/pypa/pip/blob/c247ddce123513198858d4589c5ee198dcc6ba40/src/pip/_internal/locations/_distutils.py#L9)

motor/frameworks/asyncio/__init__.py
blink1073: Yes, it hasn't been deprecated yet.  This is the workaround that tornado is using for their 6.2 release.

motor/core.py
blink1073: Done

blink1073:
We should run the synchro tests here after merging #158 to make sure there aren't any regressions.

motor/core.py
ShaneHarvey: I opened https://jira.mongodb.org/browse/MOTOR-938

motor/frameworks/asyncio/__init__.py
ShaneHarvey: Oh I see, `asyncio.set_event_loop(asyncio.new_event_loop())` will incorrectly reinitialize the loop when there is a loop already set on the thread, but it just hasn't been started yet:
```
>>> asyncio.get_running_loop()
  File "<stdin>", line 1, in <module>
RuntimeError: no running event loop
>>> id(asyncio.get_event_loop())
4361557568
>>> asyncio.set_event_loop(asyncio.new_event_loop())
>>> id(asyncio.get_event_loop())
4374658448
```

motor/core.py
blink1073: Thanks, sorry I meant to return to this comment and missed it in the scroll

pyproject.toml
blink1073: We don't need this file at all, I had added it to try to avoid the distutils deprecation warning, removing it

ShaneHarvey (approved):

motor/frameworks/asyncio/__init__.py
graingert: @blink1073 this workaround in tornado is only used for the deprecated IOLoop().start() and it is only intended to support

```
with warnings.catch_warnings():
    warnings.simplefilter("ignore")
    loop = IOLoop()
    
...
loop.start()
```

graingert: anywhere the user creates an IOLoop still happens via `asyncio.get_event_loop()` so if the user creates an IOLoop too early they get a DeprecationWarning

https://github.com/tornadoweb/tornado/blob/713f06b115f7125b4cf0d21d667ab5c801318baf/tornado/platform/asyncio.py#L295-L297

PR #156, opened 2022-04-13
MOTOR-933 Require Python 3.7+

blink1073:

PR #157, opened 2022-04-14
MOTOR-894 Snapshot Query Examples for the Manual

blink1073:


test/utils.py
```diff
@@ -127,6 +128,32 @@ def get_async_test_timeout(default=5):
         return default
 
 
+async def wait_until(predicate, success_description, timeout=10):
```
ShaneHarvey, line 131: Can we use the existing wait_until method: https://github.com/mongodb/motor/blob/6498cd5a82d50a0d81b65c9ca495dcb45838fe0c/test/py35utils.py#L20

test/asyncio_tests/test_examples.py
```diff
@@ -1302,3 +1303,105 @@ def strptime(s):
...
+            async for doc in cursor:
+                adoptablePetsCount = doc["adoptableCatsCount"]
+                break
+
+            cursor = db.dogs.aggregate(
+                [{"$match": {"adoptable": True}}, {"$count": "adoptableDogsCount"}], session=s
+            )
+            async for doc in cursor:
+                adoptablePetsCount += doc["adoptableDogsCount"]
+                break
```
ShaneHarvey, line 1344: I think it would be more idiomatic to use `to_list` for these:
```
            docs = await db.cats.aggregate(
                [{"$match": {"adoptable": True}}, {"$count": "adoptableCatsCount"}], session=s
            ).to_list(None)
            adoptablePetsCount += docs[0]["adoptableCatsCount"]
            ...
```

ShaneHarvey (changes requested):

test/utils.py
blink1073: Ha, that function actually got moved anyway as part of #156

test/asyncio_tests/test_examples.py
blink1073: Done

ShaneHarvey (approved):

PR #158, opened 2022-04-18
MOTOR-900 Fix synchro failures on PyMongo 4.1.1

blink1073:


blink1073:
All green including all syncho tests!

motor/core.py
```diff
@@ -105,6 +105,8 @@ class AgnosticClient(AgnosticBaseProperties):
     arbiters = ReadOnlyProperty()
     close = DelegateMethod()
     __hash__ = DelegateMethod()
+    __next__ = DelegateMethod()
+    next = DelegateMethod()
```
ShaneHarvey, line 109: Do we need these methods on Client/DB/Coll? I'd prefer _not_ to add the to motor unless we absolutely need to. Eventually the plan is to remove them from pymongo in 5.0 so I don't see a reason to add them to motor.

motor/core.py
```diff
@@ -810,7 +836,7 @@ def __getitem__(self, name):
         )
 
         return collection_class(
-            self.database, self.name + "." + name, _delegate=self.delegate[name]
+            self.database, self.name + "." + str(name), _delegate=self.delegate[name]
```
ShaneHarvey, line 839: I don't think we want to call str(name) since name should already be a str.

motor/motor_gridfs.py
```diff
@@ -222,6 +222,7 @@ class AgnosticGridIn(object):
     readable = DelegateMethod()
     seekable = DelegateMethod()
     upload_date = ReadOnlyProperty()
+    md5 = ReadOnlyProperty()
```
ShaneHarvey, line 225: Let's not add this property since it's useless in motor.

synchro/__init__.py
```diff
@@ -483,7 +495,7 @@ def __getattr__(self, name):
 
     def __getitem__(self, name):
         # Access to collections with dotted names, like db.test['mike']
-        fullname = self.name + "." + name
+        fullname = self.name + "." + str(name)
```
ShaneHarvey, line 498: I don't think we want to call str(name) since name should already be a str.

ShaneHarvey (changes requested):

motor/core.py
blink1073: Sounds good, I should be able to skip the relevant tests instead

motor/core.py
blink1073: At least one of the tests was passing in an integer for the name

motor/motor_gridfs.py
blink1073: Ack, I'll find a way around this, might mean skipping a couple more tests.

synchro/__init__.py
blink1073: Same comment as the core class

juliusgeo (approved):
LGTM!

motor/core.py
ShaneHarvey: We should skip those tests. Motor should not accept non-strings for getitem even if pymongo does.

synchro/__init__.py
ShaneHarvey: We should skip those tests. Motor should not accept non-strings for getitem even if pymongo does.

ShaneHarvey (changes requested):

motor/core.py
```diff
@@ -511,6 +518,7 @@ class AgnosticDatabase(AgnosticBaseProperties):
     __delegate_class__ = Database
 
     __hash__ = DelegateMethod()
+    __bool__ = DelegateMethod()
```
ShaneHarvey, line 521: This implements the breaking change in PYTHON-808 for Motor 3.0, right? If so, could you make a note about this in the changelog?

blink1073: Done

ShaneHarvey (changes requested):
Still waiting on the getitem str() issues.

ShaneHarvey (approved):

PR #160, opened 2022-04-21
MOTOR-843 Add support for the comment field to all helpers

blink1073:
Some of these were added as part of #158

### DIFF
--- a/motor/docstrings.py
+++ b/motor/docstrings.py
@@ -83,6 +83,10 @@ based only on the URI in a configuration file.
     :class:`~pymongo.read_concern.ReadConcern`. If ``None`` (the
     default) the :attr:`read_concern` of this :class:`MotorClient` is
     used.
+  - `comment` (optional): A user-provided comment to attach to this command.
+
+.. versionchanged:: 1.2
+   Added ``comment`` parameter.
 
 .. versionadded:: 2.1
    Revived this method. Added the ``default``, ``codec_options``,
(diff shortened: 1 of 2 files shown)


juliusgeo:
One thing that could be added is unit tests for the docstrings, similar to my "comment" field PR: https://github.com/juliusgeo/mongo-python-driver/blob/5160ae8da3e802baae44e2a0a5bb84e346cd67be/test/test_comment.py on line 71.

juliusgeo:
One thing that could be added is unit tests for the docstrings, similar to my "comment" field PR: https://github.com/juliusgeo/mongo-python-driver/blob/5160ae8da3e802baae44e2a0a5bb84e346cd67be/test/test_comment.py on line 71.

blink1073:
We had to skip that test in #158 because it is checking the signature, and we would have had to update all of the signatures in Motor to be explicit instead of using args and kwargs.

motor/docstrings.py
```diff
@@ -662,6 +696,9 @@ async def create_x_index():
   {'_id_': {'key': [('_id', 1)]},
    'x_1': {'unique': True, 'key': [('x', 1)]}}
 
+
```
juliusgeo, line 699: I think this blank line should be placed below.

blink1073: Done

juliusgeo (approved):
LGTM!

ShaneHarvey (approved):

PR #163, opened 2022-04-25
MOTOR-938 Docs for watch() incorrectly call ChangeStream.close()

blink1073:


doc/examples/tornado_change_stream_example.py
```diff
@@ -109,10 +109,8 @@ def main():
     try:
         loop.start()
     except KeyboardInterrupt:
-        pass
-    finally:
         if change_stream is not None:
-            change_stream.close()
+            loop.add_callback(change_stream.close)
```
ShaneHarvey, line 113: Does this actually work? I would expect close to never run since the loop is not running at this point. What if we use loop.run_sync similar to asyncio.run?:

```
    loop = tornado.ioloop.IOLoop.current()
    # Start watching collection for changes.
    try:
        loop.run_sync(lambda: watch(collection))
    except KeyboardInterrupt:
        pass
```

ShaneHarvey (changes requested):

motor/core.py
```diff
@@ -1069,19 +1069,16 @@ def main():
...
           # asyncio
           try:
-              asyncio.run(watch_collection)
+              asyncio.run(watch_collection())
           except KeyboardInterrupt:
-                  pass
-          finally:
               if change_stream is not None:
-                  change_stream.close()
+                  asyncio.run(change_stream.close())
```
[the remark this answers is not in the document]
blink1073: I did some local testing, ensuring that the change_stream was actually closed.  `asyncio.run` properly runs `__aexit__()`, closing the stream.  For the tornado loop, I had to manually run the close operation.

ShaneHarvey (approved):

motor/core.py
```diff
@@ -1065,23 +1065,17 @@ async def watch_collection():
...
           # asyncio
           try:
-              asyncio.run(watch_collection)
+              asyncio.run(watch_collection())
           except KeyboardInterrupt:
-                  pass
-          finally:
-              if change_stream is not None:
-                  change_stream.close()
+              pass
```
juliusgeo, line 1078: Does this pass need to be here?

blink1073: Yes, that or a comment, or it would be a syntax error.

juliusgeo (approved):
LGTM!

PR #165, opened 2022-04-25
MOTOR-608 Update outdated API docs

blink1073:


motor/docstrings.py
```diff
@@ -201,6 +201,82 @@ async def modify_data():
...
+    option is silently ignored by the server and unique index builds
+    using the option will fail if a duplicate value is detected.
+
+  .. note:: `partialFilterExpression` requires server version **>= 3.2**
+
+  .. note:: The :attr:`~pymongo.collection.Collection.write_concern` of
+     this collection is automatically applied to this operation.
+
+  :Parameters:
+    - `keys`: a single key or a list of (key, direction)
```
ShaneHarvey, line 269: Missing session and comment parameters.

motor/docstrings.py
```diff
@@ -1219,3 +1295,285 @@ async def coro():
         encrypted = await client_encryption.encrypt(value, ...)
         decrypted = await client_encryption.decrypt(encrypted)
 """
+
+gridfs_delete_doc = """Delete a file's metadata and data chunks from a GridFS bucket::
```
ShaneHarvey, line 1299: Let's add the session parameter to these gridfs docstrings.

ShaneHarvey (changes requested):
Nice! A few small suggestions.

motor/docstrings.py
blink1073: Done

motor/docstrings.py
blink1073: Done

ShaneHarvey (approved):

PR #167, opened 2022-04-25
MOTOR-944 Remove Deprecated APIs

blink1073:


ShaneHarvey:
We are not making these changes in Motor 3.0.

PR #168, opened 2022-04-25
MOTOR-946 Deprecated APIs will not be Removed in Motor 3.0

blink1073:


motor/motor_gridfs.py
```diff
@@ -42,6 +42,13 @@ class AgnosticGridOutCursor(AgnosticCursor):
 
     def next_object(self):
         """**DEPRECATED** - Get next GridOut object from cursor."""
+        warnings.warn(
```
ShaneHarvey, line 45: Won't this raise two duplicate warnings now?

ShaneHarvey (changes requested):

motor/motor_gridfs.py
blink1073: Good catch, added a comment instead.

ShaneHarvey (approved):

PR #172, opened 2022-08-06
MOTOR-1014 Add __aenter__ and __aexit__ for AgnosticBaseCursor so that cursor can be instantiated using the context manager

coderk17:
When I use the following code:
```python
async with self.coll.find() as cursor:
    async for record in cursor:
        yield record
```
It's going to throw an AttributeError
So I added the asynchronous context manager method
so that I did not have to close the cursor when I was done with it

### DIFF
--- a/motor/core.py
+++ b/motor/core.py
@@ -1252,6 +1252,13 @@ class AgnosticBaseCursor(AgnosticBase):
 
     __anext__ = next
 
+    async def __aenter__(self):
+        return self
+
+    async def __aexit__(self, exc_type, exc_val, exc_tb):
+        if self.delegate:
+            await self.close()
+
     def _get_more(self):
         """Initial query or getMore. Returns a Future."""
         if not self.alive:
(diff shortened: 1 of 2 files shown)


ShaneHarvey:
This seems reasonable to me although I'd like to figure why it wasn't added to begin with. @coderk17 could you add some tests for find() and aggregate() to test/asyncio_tests/test_asyncio_cursor.py?

Opened [MOTOR-1014](https://jira.mongodb.org/browse/MOTOR-1014) to track.

coderk17:
Sure, I'm glad to add these unittests.

coderk17:
> This seems reasonable to me although I'd like to figure why it wasn't added to begin with. @coderk17 could you add some tests for find() and aggregate() to test/asyncio_tests/test_asyncio_cursor.py?
> 
> Opened [MOTOR-1014](https://jira.mongodb.org/browse/MOTOR-1014) to track.

Hello, unit tests have been added and passed. @ShaneHarvey

ShaneHarvey (approved):
Lgtm

PR #177, opened 2022-08-25
MOTOR-1020 Error in Motor Synchro GridFS Tests "test_exception_file_non_existence"

juliusgeo:


blink1073 (approved):
LGTM!

blink1073:
Do we have a ticket for this failure?

```
 [2022/08/25 16:26:02.223] AssertionError: Unused test patterns: {'*.test_discovery_and_monitoring_integration_pool_cleared_error_PoolClearedError_does_not_mark_server_unknown', '*.test_spec_kmipKMS_Insert_a_document_with_auto_encryption_using_KMIP_KMS_provider', '*.test_discovery_and_monitoring_integration_find_shutdown_error_Concurrent_shutdown_error_on_find', '*.test_discovery_and_monitoring_integration_insert_shutdown_error_Concurrent_shutdown_error_on_insert'}
```

motor/motor_gridfs.py
```diff
@@ -308,7 +308,11 @@ async def __aenter__(self):
         return self
 
     async def __aexit__(self, exc_type, exc_val, exc_tb):
-        await self.close()
+        if exc_type is None:
+            await self.close()
+        else:
+            object.__setattr__(self, "_closed", True)
+        return False
```
ShaneHarvey, line 315: Can with directly call `GridIn.__exit__` instead of duplicating it?
```
    _exit = AsyncCommand('__exit__')
....
```

synchro/__init__.py
```diff
@@ -668,7 +675,11 @@ def __enter__(self):
         return self
 
     def __exit__(self, exc_type, exc_val, exc_tb):
-        self.close()
+        if exc_type is None:
+            self.close()
+        else:
+            object.__setattr__(self, "_closed", True)
+        return False
```
ShaneHarvey, line 682: We should try to rely on the motor method here otherwise we're testing synchro itself and not actually testing motor. Perhaps:
```
return self.synchronize(self.delegate.__aexit__)()
```

tox.ini
```diff
@@ -111,6 +111,7 @@ setenv =
     PYTHONPATH = {envtmpdir}/mongo-python-driver
 commands =
     git clone --depth 1 --branch master https://github.com/mongodb/mongo-python-driver.git {envtmpdir}/mongo-python-driver
+    pip install -e {envtmpdir}/mongo-python-driver
```
ShaneHarvey, line 114: I remember this caused some problems when we were on a call. Let's remove it?

ShaneHarvey (changes requested):
Feel free to remove the unused test patterns in this PR.

tox.ini
blink1073: Yeah, my impression was that this line should be there as well.

ShaneHarvey (approved):
LGTM!

PR #186, opened 2022-10-25
MOTOR-1059 Single Source the Package Version

blink1073:


ShaneHarvey (changes requested):
Missing the setup.py changes?

blink1073:
> Missing the setup.py changes?

Fixed

motor/_version.py
```diff
@@ -0,0 +1,24 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""Version-related data for motor."""
+
+version_tuple = (3, 1, 2, ".dev0")
```
ShaneHarvey, line 17: `".dev0"` -> `"dev0"`

otherwise version will be '3.1.2..dev0'

ShaneHarvey (changes requested):

motor/_version.py
ShaneHarvey: Or we could fix get_version_string to avoid that by removing leading `.`'s or something.

blink1073: I updated it to account for "rc" releases.

ShaneHarvey: How bout we copy what pymongo does to be consistent: https://github.com/mongodb/mongo-python-driver/blob/1d117c1f39983ec88376816c7e3fd73fb140b863/pymongo/_version.py#L21-L24

Pymongo's version is more robust, eg it handles `version_tuple = (3, 1, ".dev0")`

ShaneHarvey (changes requested):

blink1073:
> How bout we copy what pymongo does to be consistent

Done

ShaneHarvey (approved):

PR #187, opened 2022-11-14
MOTOR-923 Language specific examples for AWS Lambda

juliusgeo:


test/asyncio_tests/test_asyncio_ssl.py
```diff
@@ -195,3 +195,35 @@ async def test_cert_ssl_validation_hostname_fail(self):
...
+    async def test_shared_client(self):
+        # Start AWS Lambda Example 1
+        import os
+
+        from motor import MotorClient
+
+        client = MotorClient(host=os.environ["MONGODB_URI"])
+
+        def lambda_handler(event, context):
+            return client.db.command("ping")
```
blink1073, line 209: `command` is async, so these functions should be run as `asyncio.run()`, or `IOLoop.run_sync` for `tornado`.

blink1073 (changes requested):

test/tornado_tests/test_motor_auth.py
```diff
@@ -92,3 +92,37 @@ async def test_saslprep(self):
         )
 
         await client.scramtestdb.collection.insert_one({})
+
+
+class MotorAWSLambdaExamples(MotorAuthTest):
+    @gen_test
+    async def test_shared_client(self):
```
ShaneHarvey, line 99: Let's only add this for asyncio, not tornado.

test/asyncio_tests/test_asyncio_ssl.py
```diff
@@ -195,3 +195,35 @@ async def test_cert_ssl_validation_hostname_fail(self):
                 )
 
                 await client.db.collection.find_one()
+
+    @asyncio_test
+    async def test_shared_client(self):
```
ShaneHarvey, line 200: Let's move these tests to test_examples.py

ShaneHarvey (changes requested):

test/asyncio_tests/test_examples.py
```diff
@@ -1503,3 +1503,43 @@ def MongoClient(**kwargs):
...
         await client_encryption.close()
+
+
+class MotorAWSLambdaExamples(AsyncIOTestCase):
+    @asyncio_test
+    async def test_shared_client(self):
+        # Start AWS Lambda Example 1
+        import os
+
+        from motor import MotorClient
```
blink1073, line 1514: This should be `from motor.motor_asyncio import AsyncIOMotorClient` since we're targeting the `asyncio` client.

test/asyncio_tests/test_examples.py
```diff
@@ -1503,3 +1503,43 @@ def MongoClient(**kwargs):
...
+
+        client = MotorClient(host=os.environ.get("MONGODB_URI", "localhost"))
```
ShaneHarvey, line 1516: Better to set a default for MONGODB_URI before the example starts so users don't mistakenly forget to set this:
```python
        os.environ.setdefault("MONGODB_URI", "localhost")
        # Start AWS Lambda Example 2
        import os
```

ShaneHarvey (changes requested):

test/asyncio_tests/test_examples.py
```diff
@@ -1503,3 +1503,43 @@ def MongoClient(**kwargs):
...
+        async def async_handler(event, context):
+            return await client.db.command("ping")
+
+        def lambda_handler(event, context):
+            return asyncio.run(async_handler(event, context))
+
+        # End AWS Lambda Example 1
+
+    @asyncio_test
+    async def test_IAM_auth(self):
```
ShaneHarvey, line 1527: This one will need to be skipped or else the MONGODB-AWS auth step will fail.

test/asyncio_tests/test_examples.py
```diff
@@ -1503,3 +1504,46 @@ def MongoClient(**kwargs):
...
+        async def async_handler(event, context):
+            return await client.db.command("ping")
+
+        def lambda_handler(event, context):
+            return asyncio.run(async_handler(event, context))
+
+        # End AWS Lambda Example 1
+
+    @unittest.skip
+    @asyncio_test
```
blink1073, line 1529: I think you also need to remove the `asyncio_test` decorator to fix the build failure.

test/asyncio_tests/test_examples.py
```diff
@@ -1503,3 +1504,45 @@ def MongoClient(**kwargs):
...
+
+        async def async_handler(event, context):
+            return await client.db.command("ping")
+
+        def lambda_handler(event, context):
+            return asyncio.run(async_handler(event, context))
+
+        # End AWS Lambda Example 1
+
+    async def test_IAM_auth(self):
```
blink1073, line 1528: Since we're skipping the test you can remove the outer `async` to fix the build failure

test/asyncio_tests/test_examples.py
```diff
@@ -1503,3 +1504,45 @@ def MongoClient(**kwargs):
...
+class MotorAWSLambdaExamples(AsyncIOTestCase):
+    @asyncio_test
+    async def test_shared_client(self):
+        environ.setdefault("MONGODB_URI", "localhost")
+        # Start AWS Lambda Example 1
+        import os
+
+        from motor.motor_asyncio import AsyncIOMotorClient
+
+        client = AsyncIOMotorClient(host=os.environ.get("MONGODB_URI", "localhost"))
```
ShaneHarvey, line 1518: This should be `AsyncIOMotorClient(os.environ["MONGODB_URI"])`

test/asyncio_tests/test_examples.py
```diff
@@ -1503,3 +1504,45 @@ def MongoClient(**kwargs):
...
 
         await client_encryption.close()
+
+
+class MotorAWSLambdaExamples(AsyncIOTestCase):
+    @asyncio_test
+    async def test_shared_client(self):
+        environ.setdefault("MONGODB_URI", "localhost")
+        # Start AWS Lambda Example 1
+        import os
```
ShaneHarvey, line 1514: missing asyncio import.

test/asyncio_tests/test_examples.py
```diff
@@ -1503,3 +1504,45 @@ def MongoClient(**kwargs):
...
+        def lambda_handler(event, context):
+            return asyncio.run(async_handler(event, context))
+
+        # End AWS Lambda Example 1
+
+    async def test_IAM_auth(self):
+        raise unittest.skip("This test needs to be run with valid IAM credentials.")
+        environ.setdefault("MONGODB_URI", "localhost")
+        # Start AWS Lambda Example 2
+        import os
```
ShaneHarvey, line 1532: missing asyncio import.

test/asyncio_tests/test_examples.py
```diff
@@ -1503,3 +1504,45 @@ def MongoClient(**kwargs):
...
+
+        from motor.motor_asyncio import AsyncIOMotorClient
+
+        client = AsyncIOMotorClient(host=os.environ.get("MONGODB_URI", "localhost"))
+
+        async def async_handler(event, context):
+            return await client.db.command("ping")
+
+        def lambda_handler(event, context):
+            return asyncio.run(async_handler(event, context))
```
ShaneHarvey, line 1524: asyncio.run does not work here:
```python
$ cat example.py 
import asyncio

client = AsyncIOMotorClient()

async def async_handler(event, context):
    return await client.db.command("ping")

def lambda_handler(event, context):
    return asyncio.run(async_handler(event, context))

lambda_handler('event', {})
lambda_handler('event', {})
lambda_handler('event', {})
```
output:
```python
$ python example.py 
  File "example.py", line 13, in <module>
    lambda_handler('event', {})
  File "example.py", line 10, in lambda_handler
    return asyncio.run(async_handler(event, context))
  File "/Library/Developer/CommandLineTools/Library/Frameworks/Python3.framework/Versions/3.8/lib/python3.8/asyncio/runners.py", line 44, in run
    return loop.run_until_complete(main)
  File "/Library/Developer/CommandLineTools/Library/Frameworks/Python3.framework/Versions/3.8/lib/python3.8/asyncio/base_events.py", line 616, in run_until_complete
    return future.result()
  File "example.py", line 7, in async_handler
    return await client.db.command("ping")
  File "/Users/shane/git/motor/motor/metaprogramming.py", line 73, in method
    return framework.run_on_executor(
  File "/Users/shane/git/motor/motor/frameworks/asyncio/__init__.py", line 74, in run_on_executor
    return loop.run_in_executor(_EXECUTOR, functools.partial(fn, *args, **kwargs))
  File "/Library/Developer/CommandLineTools/Library/Frameworks/Python3.framework/Versions/3.8/lib/python3.8/asyncio/base_events.py", line 774, in run_in_executor
    self._check_closed()
  File "/Library/Developer/CommandLineTools/Library/Frameworks/Python3.framework/Versions/3.8/lib/python3.8/asyncio/base_events.py", line 508, in _check_closed
    raise RuntimeError('Event loop is closed')
RuntimeError: Event loop is closed
```

ShaneHarvey (changes requested):

test/asyncio_tests/test_examples.py
ShaneHarvey: Motor clients bind to the current event loop and need to keep using that same loop. asyncio.run creates, runs, and then closes a loop which is incompatible with this approach. The only way to use asyncio.run would be to also create a new client for each handler call which we don't want to do since clients are expensive and should be reused across calls. You'll need to create one loop and reuse it to synchronize all calls.

ShaneHarvey: Your approach suffers from the get_event_loop() deprecation problem:
```python
>>> asyncio.get_event_loop()
<stdin>:1: DeprecationWarning: There is no current event loop
<_UnixSelectorEventLoop running=False closed=False debug=False>
```

blink1073 (approved):
LGTM!

ShaneHarvey (approved):

blink1073 (approved):
LGTM!

ShaneHarvey (approved):

ShaneHarvey:
Could you open a new issue for the tornado install failure?

PR #189, opened 2022-12-09
MOTOR-1072 Test_shared_client failing on multiple variants

juliusgeo:


test/asyncio_tests/__init__.py
```diff
@@ -101,7 +101,8 @@ def get_client_kwargs(self, **kwargs):
             kwargs.setdefault("tlsCertificateKeyFile", CLIENT_PEM)
 
         kwargs.setdefault("tls", env.mongod_started_with_ssl)
-        kwargs.setdefault("io_loop", self.loop)
+        if not kwargs.pop("io_loop", None):
+            kwargs.setdefault("io_loop", self.loop)
```
ShaneHarvey, line 105: Why pop the io_loop kwarg here? It seems wrong to discard it.

ShaneHarvey (changes requested):

blink1073 (approved):

test/asyncio_tests/__init__.py
```diff
@@ -95,23 +95,31 @@ def setUp(self):
...
+
     def asyncio_client(self, uri=None, *args, **kwargs):
         """Get an AsyncIOMotorClient.
 
         Ignores self.ssl, you must pass 'ssl' argument.
         """
+        if kwargs.get("host", None):
+            kwargs = self._get_ssl_kwargs(**kwargs)
+        else:
+            kwargs = self.get_client_kwargs(**kwargs)
```
ShaneHarvey, line 120: It does not make sense to base this if/else on the presence of the "host" kwarg. Let's always use get_client_kwargs here and figure out a more straightforward way to avoid adding the loop parameter in for test_shared_client.

ShaneHarvey (changes requested):

test/asyncio_tests/__init__.py
```diff
@@ -95,23 +95,31 @@ def setUp(self):
...
 
+    def _get_io_loop_kwargs(self, **kwargs):
+        kwargs.setdefault("io_loop", self.loop)
         return kwargs
 
-    def asyncio_client(self, uri=None, *args, **kwargs):
+    def get_client_kwargs(self, set_loop=True, **kwargs):
+        if set_loop:
+            kwargs = self._get_io_loop_kwargs(**kwargs)
+        return self._get_ssl_kwargs(**kwargs)
```
ShaneHarvey, line 112: Now that we have set_loop, can we go back to using a single get_client_kwargs method?

ShaneHarvey (changes requested):

ShaneHarvey (approved):

PR #194, opened 2023-02-14
MOTOR-1095 Make Rust Compiler available for Cryptography build

blink1073:


.evergreen/run-tox.sh
```diff
@@ -39,7 +39,7 @@ if [ "${INSTALL_TOX}" = "true" ]; then
         source motorenv/Scripts/activate
     fi
     set -o xtrace
-    pip install tox>=3.18
+    pip install "tox>=3.18,<4"
```
juliusgeo, line 42: Why make this change?

blink1073:
```
[2023/02/14 17:07:14.554] tornado6-py38: failed with env name tornado6-py38 conflicting with base python c:\python\Python38\python.exe
 [2023/02/14 17:07:14.561]   tornado6-py38: FAIL code 1 (0.02 seconds)
```

juliusgeo: Ah, does tox 4.0+ not support Python 3.8?

.evergreen/config.yml
```diff
@@ -51,6 +51,11 @@ functions:
               export PROJECT_DIRECTORY=$(cygpath -m $PROJECT_DIRECTORY)
            fi
 
+           # Installation of cryptography requires a rust compiler on some machines
+           curl https://sh.rustup.rs -sSf | sh -s -- -y --no-modify-path
+           export PATH="$HOME/.cargo/bin:$PATH"
+           rustup toolchain install nightly -c rustc
```
ShaneHarvey, line 57: Can we use the existing rust toolchain like we do in PyMongo instead of installing it? See https://github.com/mongodb/mongo-python-driver/blob/a4c90ae157ffcb1d4a073ceeb9177400126ad871/.evergreen/config.yml#L77-L78

ShaneHarvey (changes requested):

.evergreen/run-tox.sh
blink1073: Let me try removing it again, tox 4 came out in December, we most likely would have seen this before now.

.evergreen/config.yml
blink1073: There was no `/home/admin`, nor `/home/ubuntu/.cargo` on the host, so I followed what is done in [`bson-rust`](https://github.com/mongodb/bson-rust/blob/79ab64ec484124730027c7a6cb84dd33ea0114f1/.evergreen/install-dependencies.sh#L4).

.evergreen/run-tox.sh
blink1073: Ah, it appears to be a case sensitivity issue on Windows: `Python3.8` ~= `python 3.8` in our `basepython` config.

blink1073: I believe this was caused by https://github.com/tox-dev/tox/pull/2824, which was released on 5 Jan and first started affecting our waterfall on 11 Jan.

.evergreen/config.yml
ShaneHarvey: Okay then SGTM.

ShaneHarvey (approved):

blink1073:
Done: https://jira.mongodb.org/browse/MOTOR-1098

PR #195, opened 2023-03-01
MOTOR-969 Support for Range Indexes

juliusgeo:


test/tornado_tests/test_motor_collection.py
```diff
@@ -283,6 +285,57 @@ async def test_async_create_encrypted_collection(self):
             self.addCleanup(self.db.drop_collection, "testing1", encrypted_fields=ef)
             self.assertEqual(exc.exception.code, 121)
 
+    @gen_test
+    async def test_async_encrypt_expression(self):
```
blink1073, line 289: This test and the `test_async_create_encrypted_collection` should also be in `tests/asyncio_tests/test_asyncio_collection.py`

motor/metaprogramming.py
```diff
@@ -149,6 +149,8 @@ def __init__(self, attr_name, doc=None):
 
     def create_attribute(self, cls, attr_name):
         name = self.attr_name or attr_name
+        if name == "encrypt_expression":
+            return
```
blink1073, line 153: You could instead add `None` as the last arg to `getattr()` and return if it is None.

juliusgeo: My concern with doing it that way is that we might forget to get rid of this workaround later. In general I think we do want to raise an error in this case (for example `Sync("method_that_is_mispelled"`), so I want it to be as specific as possible.

blink1073: This unconditionally gets rid of the attribute though right?  I think we need to first check if it exists, and then check if it is in an expected list of ones that may not exist yet.

juliusgeo: Ah, that is a very good point. Done.

ShaneHarvey (changes requested):

test/tornado_tests/test_motor_collection.py
juliusgeo: Done.

motor/core.py
```diff
@@ -1979,6 +1979,7 @@ class AgnosticClientEncryption(AgnosticBase):
     add_key_alt_name = AsyncCommand()
     get_key_by_alt_name = AsyncCommand()
     remove_key_alt_name = AsyncCommand()
+    encrypt_expression = AsyncCommand()
```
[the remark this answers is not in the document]
juliusgeo: That makes sense. Done.

ShaneHarvey (approved):

blink1073 (approved):

test/asyncio_tests/test_asyncio_collection.py
```diff
@@ -260,6 +266,79 @@ def test_sub_collection(self):
...
+            KMS_PROVIDERS, "keyvault.datakeys", c, CodecOptions()
+        ) as client_encryption:
+            coll, ef = await client_encryption.create_encrypted_collection(
+                database=self.db,
+                name="testing1",
+                encrypted_fields={"fields": [{"path": "ssn", "bsonType": "string", "keyId": None}]},
+                kms_provider="local",
+            )
+            with self.assertRaises(pymongo.errors.WriteError) as exc:
+                coll.insert_one({"ssn": "123-45-6789"})
```
ShaneHarvey, line 287: How is this working without the await? What is the type of `coll` here?

ShaneHarvey (changes requested):

test/asyncio_tests/test_asyncio_collection.py
juliusgeo: This should be fixed now.

ShaneHarvey: How was it passing before?

juliusgeo: Before there were no awaits, so the function calls were returning immediately and not raising any exceptions.

blink1073:
Can you pull from master to pick up the `tox` fix from https://github.com/mongodb/motor/pull/194?

juliusgeo:
@blink1073 I appear to already have that commit:
 ```
(motor3.9) ➜  motor git:(MOTOR-969) git log | grep 194
    MOTOR-1095 Make Rust Compiler available for Cryptography build (#194)
```

blink1073:
Hmm, then we might need to use `ignore_basepython_conflict = true`, https://tox.wiki/en/latest/config.html#python-language-core-options

test/asyncio_tests/test_asyncio_collection.py
ShaneHarvey: Apparently this was failing before.

ShaneHarvey (approved):

juliusgeo:
@blink1073 I opened https://jira.mongodb.org/browse/MOTOR-1104 to track that work.

blink1073 (approved):

PR #196, opened 2023-03-13
MOTOR-1087 motor_asyncio freezes when using multiprocessing

juliusgeo:


blink1073 (approved):
LGTM!

motor/frameworks/asyncio/__init__.py
```diff
@@ -66,6 +66,19 @@ def get_future(loop):
...
+def _reset_global_executor():
+    """Re-initialize the global ThreadPoolExecutor"""
+    global _EXECUTOR
+    _EXECUTOR = ThreadPoolExecutor(max_workers=max_workers)
+
+
+if hasattr(os, "register_at_fork"):
+    # This will run in the same thread as the fork was called.
+    # If we fork in a critical region on the same thread, it should break.
+    # This is fine since we would never call fork directly from a critical region.
```
ShaneHarvey, line 78: What is this comment trying to say?

test/asyncio_tests/test_asyncio_basic.py
```diff
@@ -145,3 +148,28 @@ def __getitem__(self, name):
...
+                    "_EXECUTOR was not reinitialized",
+                )
+            )
+            os._exit(0)
+        else:  # Parent
+            self.assertEqual(init_exec, id(asyncio._EXECUTOR))
+            child_exec = parent_conn.recv()
+            self.assertIsNot(child_exec, init_exec)
+            passed, msg = parent_conn.recv()
+            self.assertTrue(passed, msg)
```
ShaneHarvey, line 175: Instead of testing that the id of asyncio._EXECUTOR changes, we should test that a new motor client can actually be used in the child process.

ShaneHarvey (changes requested):

motor/frameworks/asyncio/__init__.py
juliusgeo: I copied this over from Ben's PR, which had this comment for all of the register_at_fork calls. I can remove if you think it is not valuable?

ShaneHarvey: Yeah I don't think its valuable. The problem is that it's only describing _what_ after_in_child does. If there was a comment here it should explain _why_ we call it.

juliusgeo: Should be fixed now.

test/asyncio_tests/test_asyncio_basic.py
juliusgeo: Done.

blink1073 (approved):

motor/frameworks/asyncio/__init__.py
```diff
@@ -66,6 +66,17 @@ def get_future(loop):
...
 
 
+def _reset_global_executor():
+    """Re-initialize the global ThreadPoolExecutor"""
+    global _EXECUTOR
+    _EXECUTOR = ThreadPoolExecutor(max_workers=max_workers)
+
+
+if hasattr(os, "register_at_fork"):
+    # We need this to make sure that creating new clients in threads doesn't deadlock.
```
ShaneHarvey, line 76: threads -> subprocesses

ShaneHarvey (approved):

PR #197, opened 2023-03-14
Update aiohttp example

Shashankwer:
Hi,

For a newbie to motor and asyncio the code proposed here would not work especially for python3.9 onwards. asyncio automatically closes the event loop at the end of asyncio.run() method. Subsequently the aiohttp is not able to open new event loop[issue is faced on all platforms: windows, linux, mac_os]. Proposed are the changes works and ensures that the event loop is consistent among aiohttp and motor's event loop

### DIFF
--- a/doc/examples/aiohttp_example.py
+++ b/doc/examples/aiohttp_example.py
@@ -53,6 +53,6 @@ def main():
     app = init_connection()
     web.run_app(app)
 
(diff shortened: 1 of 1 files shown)


blink1073:
Thanks @Shashankwer!  Can you please put the removed comment markers back?  We need them in order for the examples page to render properly.

PR #199, opened 2023-03-21
MOTOR-1106 ERROR: module 'os' has no attribute 'fork' (AttributeError)

juliusgeo:


blink1073 (approved):
Thanks!

test/asyncio_tests/test_asyncio_basic.py
```diff
@@ -151,6 +152,7 @@ def __getitem__(self, name):
 
 
 class ExecutorForkTest(AsyncIOTestCase):
+    @unittest.skipIf(sys.platform == "win32", "This test requires fork")
```
ShaneHarvey, line 155: It would be more robust to do: `hasattr(os, "fork)`

ShaneHarvey (changes requested):

test/asyncio_tests/test_asyncio_basic.py
juliusgeo: Done.

ShaneHarvey (approved):

PR #205, opened 2023-05-11
MOTOR-1100 Update libmongocrypt payloads to new QE protocol

blink1073:


blink1073:
Failures are pre-existing and tracked in https://jira.mongodb.org/browse/MOTOR-1133

tox.ini
```diff
@@ -110,7 +110,8 @@ commands =
 
 [testenv:py3-pymongo-latest]
 commands =
-    pip install git+https://github.com/mongodb/mongo-python-driver.git@master#egg=pymongo[encryption]
+    pip install git+https://github.com/mongodb/mongo-python-driver.git@master
+    pip install --pre pymongocrypt
```
ShaneHarvey, line 114: I'm curious, why did you decide to use `--pre` even though we didn't publish a prerelease?

blink1073: I had mentioned in the ticket that this would allow us to publish a beta and consume it here.

blink1073: (in the future)

ShaneHarvey: SGTM

ShaneHarvey (approved):

PR #212, opened 2023-06-13
MOTOR-1140 Switch to using RHEL only for tests

blink1073:


.evergreen/config.yml
```diff
@@ -859,14 +859,18 @@ axes:
...
+      - id: "tornado6-py311"
+        variables:
+           TOX_ENV: "tornado6-py311"
+           PYTHON_BINARY: "/opt/python/3.11/bin/python3"
       - id: "tornado_git-py38"
         variables:
            TOX_ENV: "tornado_git-py38"
            PYTHON_BINARY: "/opt/python/3.8/bin/python3"
-      - id: "asyncio-pypy37"
+      - id: "asyncio-pypy38"
```
ShaneHarvey, line 870: Is there another platform we can test with pypy 3.7?

blink1073: Found a workaround, by adding an upper limit for that combination.

ShaneHarvey (approved):

PR #215, opened 2023-06-21
MOTOR-1144 Release 3.2

blink1073:


ShaneHarvey (commented):
Should we bump motor's min pymongo dep to >=4.4?

blink1073:
> Should we bump motor's min pymongo dep to >=4.4?

Done

ShaneHarvey (approved):

PR #217, opened 2023-07-28
MOTOR-1164 Update MongoClient import path

fergusdixon:
Pymongo [recommends](https://pymongo.readthedocs.io/en/stable/tutorial.html#making-a-connection-with-mongoclient) importing the MongoClient from the top level module

[dd-trace-py relies on this import path](https://github.com/DataDog/dd-trace-py/blob/926a210494dd6ea0f396a890bcbf6033c45d4e74/ddtrace/contrib/pymongo/patch.py#L36) to properly instrument the client in Datadog

Patching the deeper-import results in errors creating the class as can be seen [here](https://github.com/DataDog/dd-trace-py/pull/6343)

Motor should import the client from the recommended location

### DIFF
--- a/doc/contributors.rst
+++ b/doc/contributors.rst
@@ -17,3 +17,4 @@ The following is a list of people who have contributed to
 - Tushar Singh
 - Steven Silvester
 - Julius Park
+- Fergus Strangways-Dixon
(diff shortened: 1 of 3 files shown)


ShaneHarvey:
This looks like a bug in DataDog. It's just as valid to import MongoClient from pymongo and it is pymongo.mongo_client.

PR #220, opened 2023-08-09
MOTOR-1136 Ability to create and manage Atlas search indexes in the database

blink1073:


blink1073:
We could add a guard on whether `create_search_index` is defined, but the tests also rely on the new `pool.conns` API.

motor/core.py
```diff
@@ -722,6 +722,87 @@ def watch(
...
+        self,
+        command,
+        value=1,
+        read_preference=None,
+        codec_options=None,
+        session=None,
+        comment=None,
+        max_await_time_ms=None,
+        **kwargs,
+    ):
```
ShaneHarvey, line 735: Wow it feels odd to not have type annotations now.

motor/core.py
```diff
@@ -722,6 +722,87 @@ def watch(
...
+        """
+        args = (command,)
+        kwargs["value"] = value
+        kwargs["read_preference"] = read_preference
+        kwargs["codec_options"] = codec_options
+        kwargs["session"] = session
+        kwargs["comment"] = comment
+        kwargs["max_await_time_ms"] = max_await_time_ms
+
+        cursor = self.delegate.cursor_command(
```
ShaneHarvey, line 793: This needs to be run asynchronously on the background thread.

motor/core.py
```diff
@@ -1670,6 +1773,37 @@ class AgnosticCommandCursor(AgnosticBaseCursor):
 
     _CommandCursor__die = AsyncRead()
 
+    def _lazy_init(self):
+        if not self.delegate:
+            self.delegate = self._target.delegate.watch(**unwrap_kwargs_session(self._kwargs))
```
ShaneHarvey, line 1778: This is not needed for AgnosticCommandCursor.

motor/core.py
```diff
@@ -1670,6 +1773,37 @@ class AgnosticCommandCursor(AgnosticBaseCursor):
...
+        :Returns:
+          The next document or ``None`` when no document is available
+          after running a single getMore or when the cursor is closed.
+        """
+        loop = self.get_io_loop()
+        return await self._framework.run_on_executor(loop, self._try_next)
+
+    # TODO: MOTOR-1169
+    if hasattr(CommandCursor, "try_next"):
+        try_next = __try_next
```
ShaneHarvey, line 1805: Could you add a manual test for this? Eg a test that uses cursor_command and then calls try_next?

synchro/__init__.py
```diff
@@ -483,6 +484,7 @@ class Collection(Synchro):
     aggregate = WrapOutgoing()
     aggregate_raw_batches = WrapOutgoing()
     list_indexes = WrapOutgoing()
+    list_search_indexes = WrapOutgoing()
```
ShaneHarvey, line 487: We should add the other search methods too.

ShaneHarvey (changes requested):

motor/core.py
blink1073: Done

motor/core.py
blink1073: Removed

motor/core.py
blink1073: Done

synchro/__init__.py
blink1073: Done

motor/core.py
```diff
@@ -722,6 +722,92 @@ def watch(
...
+          - `read_preference` (optional): The read preference for this
+            operation. See :mod:`~pymongo.read_preferences` for options.
+            If the provided `session` is in a transaction, defaults to the
+            read preference configured for the transaction.
+            Otherwise, defaults to
+            :attr:`~pymongo.read_preferences.ReadPreference.PRIMARY`.
+          - `codec_options`: A :class:`~bson.codec_options.CodecOptions`
+            instance.
+          - `session` (optional): A
+            :class:`~pymongo.client_session.ClientSession`.
```
ShaneHarvey, line 763: Should be a motor session, not pymongo.

synchro/__init__.py
```diff
@@ -468,6 +468,17 @@ def __init__(self, client, name, **kwargs):
     def client(self):
         return self._client
 
+    def cursor_command(self, *args, **kwargs):
+        if "session" in kwargs:
+            # Workaround for validation added in PYTHON-3228.
+            kwargs["session"] = kwargs["session"].delegate
+        cursor = self.synchronize(self.delegate.cursor_command)(*args, **kwargs)
+        cmd_cursor = CommandCursor(cursor)
+        # Send the initial command as PyMongo expects.
```
ShaneHarvey, line 477: I believe this part should be removed IIUC. The PyMongo cursor_command method never runs getMore so neither should synchro.

ShaneHarvey (changes requested):

synchro/__init__.py
blink1073: There is a `test.test_run_command.TestUnifiedRunCursorCommand.test_accepts_an_explicit_session_that_is_reused_across_getMores` test.

motor/core.py
blink1073: Fixed

synchro/__init__.py
ShaneHarvey: That unified test operation works differently because it iterates the resulting cursor returned by run_command, see [here](https://github.com/mongodb/specifications/blob/master/source/unified-test-format/unified-test-format.rst#runcursorcommand):
> When executing the provided command, the test runner MUST fully iterate the cursor.

So the getMore part here still needs to be removed.

blink1073: Done

ShaneHarvey (approved):
LGTM!

PR #226, opened 2023-08-30
MOTOR-1183 Typing issue in core.pyi causing MyPy error for to_list …

imSniff:
Related with this ticket:
https://jira.mongodb.org/projects/MOTOR/issues/MOTOR-1183?filter=allopenissues
Here is the mypy error when it's using only List as a return value of the function to_list().
image


motor/core.pyi
```diff
@@ -15,6 +15,7 @@
 """Framework-agnostic type stubs for Motor, an asynchronous driver for MongoDB."""
 
 from __future__ import annotations
+from asyncio import Future
```
imSniff, line 19:
```suggestion

from asyncio import Future
```

blink1073 (approved):
Thank you!  We had missed this one and `fetch_next` because the original methods had a `coroutine_annotation` decorator instead of being `async def`.

imSniff:
> Thank you! We had missed this one and `fetch_next` because the original methods had a `coroutine_annotation` decorator instead of being `async def`.

Thank you too, for your quickly response! I just only saw this one for the moment, but if I find another typing issues, I will let you know. :)

blink1073:
I will release a bug fix today.

PR #230, opened 2023-09-06
MOTOR-1178 Stop using tox labels

blink1073:


tox.ini
NoahStapp, line 1: We should require `tox>=4.9.0` to ensure the safety of not using labels.

NoahStapp (changes requested):

tox.ini
blink1073: Done.  TIL about [Auto-Provisioning](https://tox.wiki/en/latest/user_guide.html#auto-provisioning), which works for tox 3.8.0+

blink1073: Ah, no can do, tox 4.9 only supports Python 3.8+

NoahStapp: Then all of our Python 3.7 environments could have tox commands fail silently without the labels. Maybe we delay this change until we drop support for Python 3.8?

blink1073: It would only fail silently if the label was incorrect, no?  But we're using the same labels across Python versions.

NoahStapp: But without labels you introduce the possibility of typos or other mistakes to silently fail. With labels, your environments won't run at all if you don't explicitly add a label for them.

blink1073: You're talking about for local development?

NoahStapp: By using `tox -m <label>`, we ensure that it fails instantly if the environment does not exist. Since the same behavior only exists for `tox -e <env>` in Python 3.8+, all of our Python 3.7 tests will not fail on nonexistent environment names.

blink1073: And my argument is that we should instead ensure that we are running the same code paths on Python 3.7 and a non-EOL python, thereby avoiding that risk

NoahStapp: Ah, I see sorry! Didn't understand your suggestion: is there a way of doing that without manually keeping the 3.7 version in sync?

blink1073: I went down a rabbit hole of trying to get all the tasks to run on 3.7 and 3.12, but I don't think it is worth it, let's just close these tickets.

PR #233, opened 2023-11-08
MOTOR-1209: Motor's DriverInfo should not be overwritten

Jibola:
Ticket: https://jira.mongodb.org/browse/MOTOR-1209


motor/core.py
```diff
@@ -140,9 +140,18 @@ def __init__(self, *args, **kwargs):
...
 
         kwargs.setdefault("connect", False)
-        kwargs.setdefault(
-            "driver", DriverInfo("Motor", motor_version, self._framework.platform_info())
-        )
+
+        driver_info = DriverInfo("Motor", motor_version, self._framework.platform_info())
+
+        if kwargs.get("driver"):
+            provided_info = kwargs.get("driver")
```
ShaneHarvey, line 147: We should check the type here to give a nice error message in this case `Client(driver="string")`

motor/core.py
```diff
@@ -486,7 +495,11 @@ async def coro(session, custom_arg, custom_kwarg=None):
                 return ret
 
     def start_transaction(
-        self, read_concern=None, write_concern=None, read_preference=None, max_commit_time_ms=None
+        self,
+        read_concern=None,
+        write_concern=None,
+        read_preference=None,
+        max_commit_time_ms=None,
```
ShaneHarvey, line 502: What's up with these changes?

motor/core.py
```diff
@@ -140,9 +140,18 @@ def __init__(self, *args, **kwargs):
         self._io_loop = io_loop
 
         kwargs.setdefault("connect", False)
-        kwargs.setdefault(
-            "driver", DriverInfo("Motor", motor_version, self._framework.platform_info())
-        )
+
+        driver_info = DriverInfo("Motor", motor_version, self._framework.platform_info())
+
+        if kwargs.get("driver"):
```
ShaneHarvey, line 146: Can we add a test that this is sent to pymongo correctly?

ShaneHarvey (changes requested):

motor/core.py
```diff
@@ -140,9 +140,18 @@ def __init__(self, *args, **kwargs):
...
+            provided_info = kwargs.get("driver")
+            driver_info = DriverInfo(
+                f"{driver_info.name}|{provided_info.name}",
+                f"{driver_info.version}|{provided_info.version}",
+                provided_info.platform or driver_info.platform,
```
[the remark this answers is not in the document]
Jibola: Ah you're right. I had replicated the JIRA spec and made the assumption that platform didn't get the same treatment.

motor/core.py
Jibola: I think that's the auto-formatter at work. I let it be as I assumed it came from the pre-commit hook.

Jibola: Yeah, this is a byproduct of running `tox -m lint` before I pushed.

Jibola: I've gone ahead and re-added it, but as a separate commit.

test/tornado_tests/test_motor_client.py
```diff
@@ -307,6 +308,40 @@ async def test_handshake(self):
...
+        self.assertEqual(f"PyMongo|Motor|{driver_info.name}", meta["driver"]["name"])
+        self.assertIn("Tornado", meta["platform"])
+        self.assertIn(f"|{driver_info.platform}", meta["platform"])
+        self.assertTrue(
+            meta["driver"]["version"].endswith(f"{motor.version}|{driver_info.version}"),
+            "Version in handshake [%s] doesn't end with MotorVersion|Test version [%s]"
+            % (meta["driver"]["version"], f"{motor.version}|{driver_info.version}"),
+        )
+
+        ismaster.hangs_up()
```
ShaneHarvey, line 330: Can this be `.ok()` instead?

test/tornado_tests/test_motor_client.py
```diff
@@ -307,6 +308,40 @@ async def test_handshake(self):
...
             pass
 
+    @gen_test
+    async def test_driver_info(self):
+        server = self.server()
+        driver_info = DriverInfo(name="Foo", version="1.1.1", platform="FooPlat")
+        client = motor.MotorClient(server.uri, driver=driver_info)
+
+        # Trigger connection.
+        future = client.db.command("ping")
```
ShaneHarvey, line 318: Is this needed? IIUC the ismaster is from SDAM which happens automatically so this future could be removed.

Jibola: Ah, I wasn't sure if it was needed. I assumed that the ping was the source.

test/tornado_tests/test_motor_client.py
```diff
@@ -307,6 +308,40 @@ async def test_handshake(self):
...
+        ismaster = await self.run_thread(server.receives, "ismaster")
```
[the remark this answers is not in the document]
Jibola: Yeah, agreed. Is there a task for removing the `ismaster` references in situations like this generally? I know we have one for `hello`. I checked but couldn't find one. If not I'll cut a task and wrap that up, too.

motor/core.py
```diff
@@ -140,9 +140,23 @@ def __init__(self, *args, **kwargs):
...
+            provided_info = kwargs.get("driver")
+            if not isinstance(provided_info, DriverInfo):
+                raise TypeError(
+                    f"Incorrect type for `driver` {type(provided_info)};"
+                    " expected value of type pymongo.driver_info.DriverInfo"
+                )
+            driver_info = DriverInfo(
+                f"{driver_info.name}|{provided_info.name}",
+                f"{driver_info.version}|{provided_info.version}",
+                f"{driver_info.platform}|{provided_info.platform}",
```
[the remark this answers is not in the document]
Jibola: Sounds good to me.

test/tornado_tests/test_motor_client.py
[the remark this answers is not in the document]
ShaneHarvey: We can't remove the `(server.receives, "ismaster")` part so no need to create a new ticket.

test/tornado_tests/test_motor_client.py
Jibola: Yeah, I tried removing the ping call, and the test now crashes with `asyncio.exceptions.TimeoutError: Operation timed out after 5 seconds`

ShaneHarvey (changes requested):
Could you `git merge master` to make it easier to ensure this new test passes?

ShaneHarvey (approved):
LGTM!

PR #237, opened 2023-11-10
MOTOR-1187 Fix Motor Types

blink1073:


ShaneHarvey (commented):
What about the read preference vs _ServerMode issue the user mentioned in the ticket? Can we do anything there?

> AgnosticClientSession.with_transaction read_preference parameter is typed as Optional[_ServerMode] but takes Optional[pymongo.ReadPreference] type

I'm not sure why this would be specific to with_transaction though. It looks like we use _ServerMode everywhere we expect a read_preference right now.

blink1073:
> I'm not sure why this would be specific to with_transaction though. It looks like we use _ServerMode everywhere we expect a read_preference right now.

It was already correct and we had a test for it, later in the thread the original reporter said it was a false alarm.

ShaneHarvey (approved):
Lgtm

PR #239, opened 2023-11-10
PYTHON-4040 Pin testing to aiohttp<3.8.6

Jibola:
Our tests break when we leverage `aiohttp>=3.8.6` tests will fail.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -66,7 +66,7 @@ srv = [
     "pymongo[srv]>=4.5,<5",
 ]
 test = [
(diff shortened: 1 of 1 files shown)


NoahStapp (approved):

ShaneHarvey:
Why does this test only fail when testing pymongo latest and not 4.6?

Jibola:
> Why does this test only fail when testing pymongo latest and not 4.6?

I'm not sure. One think of note is that this fails when manually running tests, but I tried using `tox` and things worked just fine.

ShaneHarvey (approved):
LGTM, let's open a new ticket to investigate the potential bug in aiohttp.

PR #240, opened 2023-11-14
MOTOR-1219 Use PyPI Trusted Publishing

blink1073:
Based on https://github.com/pypa/wheel/blob/main/.github/workflows/publish.yml

PR #241, opened 2023-11-14
MOTOR-1213 Release 3.3.2 changelog update

NoahStapp:


motor/_version.py
```diff
@@ -13,7 +13,7 @@
 # limitations under the License.
 
 """Version-related data for motor."""
-version_tuple = (3, 4, 0, ".dev0")
+version_tuple = (3, 3, 2, ".dev0")
```
blink1073, line 16: We should remove the `.dev0` portion.

blink1073 (changes requested):

blink1073 (approved):
LGTM!

motor/_version.py
ShaneHarvey: This should not have been removed until all the 3.3.2 commits are merged. Stilling waiting on https://github.com/mongodb/motor/pull/233 and https://github.com/mongodb/motor/pull/240 IIUC.

ShaneHarvey: (For next time). No need to change it now.

PR #242, opened 2023-11-21
MOTOR-1221 Add pyi lint checks

blink1073:


motor/core.pyi
```diff
@@ -572,7 +568,7 @@ class AgnosticCollection(AgnosticBaseProperties[_DocumentType]):
         collation: Optional[_CollationIn] = None,
         array_filters: Optional[Sequence[Mapping[str, Any]]] = None,
         hint: Optional[_IndexKeyHint] = None,
-        session: Union[Optional[AgnosticClientSession], Optional[AgnosticClientSession]] = None,
```
Jibola, line 575: Interesting. Why was this ever needed?

Jibola (approved):

motor/core.pyi
blink1073: It wasn't, I think it is a copy-paste artifact.

PR #243, opened 2023-11-26
MOTOR-1223 Fix motor asyncio __all__

llirrikk:


motor/motor_asyncio.py
```diff
@@ -19,7 +19,7 @@
 
 __all__ = [
     "AsyncIOMotorClient",
-    "AsyncIOMotorClientEncryption",
```
ShaneHarvey, line 22: Was AsyncIOMotorClientEncryption removed by mistake?

ShaneHarvey (changes requested):

motor/motor_asyncio.py
llirrikk: No, the last element of the `__all__` list is exactly `AsyncIOMotorClientEncryption`. The two elements (22, 32 lines) identical:
https://github.com/mongodb/motor/blob/bb1e4b9ba1dca8c0302eb3777081ea64c74d4c76/motor/motor_asyncio.py#L20-L33

ShaneHarvey (changes requested):

motor/motor_asyncio.py
llirrikk: done

ShaneHarvey (approved):

ShaneHarvey:
Thanks!

PR #244, opened 2023-12-12
MOTOR-1224 Convert top level docs to Markdown

blink1073:


ShaneHarvey (changes requested):
Do we need to update the pyproject readme too?

blink1073:
> Do we need to update the pyproject readme too?

Done!

ShaneHarvey (approved):

PR #246, opened 2023-12-22
Typehints for "to_list"

Rimuwu:
Typehints have been added so that code analyzers better understand that None can be passed to a function


blink1073:
Hi @Rimuwu, our typings file has this typed as `Optional[int]`: https://github.com/mongodb/motor/blob/fc84d28fabf6dc9fdd68ac5aca3785c78fddc407/motor/core.pyi#L682.  

Your type checker should be picking up that file.  Can you describe more your setup and what error you received?

Rimuwu:
I pass to to_list - None, which means that I do not impose restrictions. But the plugin in VS tells me that this is a mistake.

"An argument of type "None" cannot be assigned to the "length" parameter of type "int" in
the "to_list" function "None" is incompatible with "int" Pylance(reportGeneralTypeIssues)"

caseyclements:
Hi @Rimuwu ! I've recently joined MongoDB, and I had the same double-take on this usage. Might I suggest that instead of making this change in the .py file, you do it in the stub file `core.pyi`? 

`def to_list(self, length: Optional[int]) -> Future[list]: ...` -> `def to_list(self, length: Union[int, None]) -> Future[list]: ...`

blink1073:
I created https://jira.mongodb.org/browse/MOTOR-1251 for the docs failure

Rimuwu:
🎉

blink1073:
Actually we already updated the docs config in [MOTOR-1232](https://jira.mongodb.org/browse/MOTOR-1232).  I'm just waiting for CI to finish before merging.

PR #249, opened 2024-01-16
MOTOR-1235 - Fix typing for index listing

kikones34:
Type hints are currently broken for `list_search_indexes`:

Screenshot 2024-01-16 162735

The actual type returned is as follows:

Screenshot 2024-01-16 162705

Also affects `list_indexes`, although it's not as broken.

This PR fixes this.


blink1073 (approved):
Thank you!

PR #256, opened 2024-02-03
MOTOR-1245 Ensure secrets are not logged in Evergreen

blink1073:


.evergreen/config.yml
```diff
@@ -417,6 +417,7 @@ functions:
       type: test
       params:
         working_dir: "src"
+        silent: true
```
ShaneHarvey, line 420: Instead of silencing all test output, which makes it impossible to see why a run failed, could we silently load the SASL vars like we did in: https://github.com/mongodb/mongo-python-driver/pull/1072?

ShaneHarvey (changes requested):

.evergreen/config.yml
ShaneHarvey: Alternatively we can using include_expansions_in_env

blink1073: Updated to use the vault

ShaneHarvey (approved):

NoahStapp (approved):

PR #258, opened 2024-02-06
MOTOR-1248 Add Dependabot Config for GitHub Actions

blink1073:


caseyclements (approved):
Yup. Looks identical to mongo-python-driver!

PR #261, opened 2024-02-20
MOTOR-1240 Skip GridFS tests failing due to PYTHON-4146

ShaneHarvey:
MOTOR-1240

### DIFF
--- a/synchro/synchrotest.py
+++ b/synchro/synchrotest.py
@@ -131,6 +131,10 @@ excluded_tests = [
     "TestCursor.test_comment",
     "TestCursor.test_where",
     "TestGridfs.test_gridfs_find",
+    # Uses patching on grid_file._UPLOAD_BUFFER_SIZE which fails in synchro.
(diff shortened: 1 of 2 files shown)


ShaneHarvey:
The remaining test failures are tracked in other issues.

synchro/synchrotest.py
```diff
@@ -131,6 +131,10 @@
     "TestCursor.test_comment",
     "TestCursor.test_where",
     "TestGridfs.test_gridfs_find",
+    # Uses patching on grid_file._UPLOAD_BUFFER_SIZE which fails in synchro.
```
Jibola, line 134: Do we need to attach a PYTHON TODO jira ticket to this?

Jibola (approved):

synchro/synchrotest.py
ShaneHarvey: No I see it as not worth the effort.

PR #269, opened 2024-03-22
MOTOR-1279 Improve Motor Typing

blink1073:
- Updates typings for `core.pyi` to consistently return Agnostic types
- Updates `motor_asyncio.pyi` and `motor_tornado.pyi` to subclass the Agnostic types and return framework-specific types
- Updates `test/test_typing.py` to use `AsyncIOMotorClient` directly
- Adds explicit typing test to ensure `client.get_database()` and `client[database]` are identical.


ShaneHarvey (approved):
LGTM!

PR #272, opened 2024-03-23
MOTOR-1285 Add generics to type stubs

Doekeb:
This PR adds generics to type stubs to match `pymongo`. For example, `MongoClient` is generic on its `_DocumentType`. This PR makes `AgnosticClient`, `AsyncIOMotorClient`, and `MotorClient` also generic on the same `_DocumentType` (as well as all the other classes where the corresponding `pymongo` class is generic, including `Database`, `Collection`, `ChangeStream`, and the various `Cursor`s).


blink1073:
Hi @Doekeb, thanks for adding this!  Out of curiosity, were you wanting to use this feature yourself, or did you want to add this for completeness with the PyMongo API?

Also, can you please add a test similar to this one in [PyMongo](https://github.com/mongodb/mongo-python-driver/blob/42a08c4a34c1bd58296de1ffa2368396fdb12f8f/test/test_typing.py#L369)?  

Of note, we are planning to release 3.4 either tomorrow or the next day.

Doekeb:
Thanks for taking a look! I'm definitely interested in using it myself, but I do think it's quite nice to have parity with PyMongo. My use case is a data access API layer where both PyMongo and Motor are used (PyMongo for everyday use, e.g. fetch some data in a notebook and put it into a pandas DataFrame, and Motor mainly for listening to a change stream). It's nice to be able to have parallel definitions everywhere. Also, making everything generic on the document type allows one to be more sure about the nature of returned documents (e.g. you can do more things with a `dict[str, Any]` or a `RawBSONDocument` than you can with a general `Mapping[str, Any]`).

I probably won't have a chance to implement those test for at least the next couple of days, but I'm happy to do so when I get a chance!

blink1073:
Okay, sounds good!  We'll save it for the 3.5 release then, which will most likely coincide with PyMongo 4.7.

Doekeb:
Just added those tests, ready for another look!

blink1073 (approved):
Thanks very much!

PR #283, opened 2024-06-03
MOTOR-1299 Add SBOM-lite and CodeQL files

blink1073:


blink1073:
I opened https://jira.mongodb.org/browse/MOTOR-1322 to track the failures which were caused by an upstream change in PyMongo.

caseyclements:
So, the failures are red herrings, and this is good to go?

blink1073:
> So, the failures are red herrings, and this is good to go?

Yes, but I just pulled in #284 to verify.

caseyclements (approved):
LGTM

PR #297, opened 2024-08-13
MOTOR-1352 Update APIs to Support PyMongo 4.9

blink1073:
Synchro tests updates are deferred to [MOTOR-1353](https://jira.mongodb.org/browse/MOTOR-1353).
Docs tests are also expected to fail until https://jira.mongodb.org/browse/MOTOR-1351


motor/core.py
```diff
@@ -1769,7 +1770,7 @@ class AgnosticCursor(AgnosticBaseCursor):
     comment = MotorCursorChainingMethod()
     allow_disk_use = MotorCursorChainingMethod()
 
-    _Cursor__die = AsyncRead()
+    _die_lock = AsyncRead()
```
ShaneHarvey, line 1773: Do we even need _die_lock defined anymore?

blink1073: Just in `_LatentCursor` it seems.  I removed the others.

ShaneHarvey: Yeah the _LatentCursor behavior is odd. That's another reason I want to avoid implementing this pattern in pymongo.

ShaneHarvey (approved):

PR #299, opened 2024-08-19
MOTOR-1358 Fix AgnosticGridFSBucket collection type

jochman:
As "collection" is just an alias to "bucket_name", it should be str.
if it's of type collection - it fails in code as it tries to use it as a string.

from motor.motor_gridfs.py:
```python
        if collection is not None:
            warnings.warn(
                'the "collection" parameter is deprecated, use "bucket_name" instead',
                DeprecationWarning,
                stacklevel=2,
            )
            bucket_name = collection
```

`motor/motor/motor_gridfs.py:340`: 
```python
class AgnosticGridFSBucket:
    def __init__(
        self,
        database,
        bucket_name="fs",
        chunk_size_bytes=DEFAULT_CHUNK_SIZE,
        write_concern=None,
        read_preference=None,
        collection=None,
    ):...
```


ShaneHarvey (approved):
LGTM!

PR #305, opened 2024-09-09
MOTOR-1363 - Pass correct parameter to AsyncIOMotorClientEncryption

NoahStapp:


NoahStapp:
The tests appear to be failing now due to not closing clients correctly--do we want to wait for that ticket to be done first, or have this PR also skip those warnings for now?

blink1073:
> have this PR also skip those warnings for now?

I'd say skip them.  There is a separate set of failures in synchro though

synchro/synchrotest.py
```diff
@@ -334,17 +334,8 @@ def want_method(method, classname):
...
         "pymongo.database",
         "pymongo.srv_resolver",
-        "pymongo.synchronous.collection",
-        "pymongo.synchronous.client_session",
-        "pymongo.synchronous.command_cursor",
-        "pymongo.synchronous.change_stream",
-        "pymongo.synchronous.cursor",
-        "pymongo.synchronous.encryption",
-        "pymongo.synchronous.mongo_client",
-        "pymongo.synchronous.database",
```
ShaneHarvey, line 344: Could you explain these changes?

ShaneHarvey: Ah this is exactly the case I was worried about here: https://github.com/mongodb/mongo-python-driver/pull/1820#discussion_r1739317264

NoahStapp: A very accurate prediction, unfortunately I can't find a solution for this issue in Motor that doesn't rely on also changing PyMongo code.

blink1073: an option is to get the class name of the full mro and check for the desired name

NoahStapp: Something like `type(client).__name__ == "MongoClient"`?

blink1073: But also for subclasses:

```
>>> class Foo(MongoClient):
...    pass
...
>>> Foo.__mro__
(<class '__main__.Foo'>, <class 'pymongo.synchronous.mongo_client.MongoClient'>, <class 'pymongo.common.BaseObject'>, <class 'typing.Generic'>, <class 'object'>)
```

NoahStapp: We'd still run into the same issue with that, right? Since the import paths differ.

blink1073 (approved):
LGTM!

PR #311, opened 2024-09-30
DEVPROD-11430 Use temporary aws creds for s3 uploads

blink1073:
Also removes unused functions that would have required elevated permissions.


blink1073:
evergreen retry

PR #312, opened 2024-11-11
PYTHON-4962 Adopt zizmor GitHub Actions security scanner

blink1073:


blink1073:
I opened https://jira.mongodb.org/browse/MOTOR-1413 to track the failures.

NoahStapp (approved):

.github/workflows/zizmor.yml
```diff
@@ -0,0 +1,32 @@
+name: GitHub Actions Security Analysis with zizmor 🌈
```
NoahStapp, line 1: Is this 🌈 emoji actually part of the name?

.github/workflows/zizmor.yml
```diff
@@ -0,0 +1,32 @@
...
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v4
+        with:
+          persist-credentials: false
+      - name: Setup Rust
+        uses: actions-rust-lang/setup-rust-toolchain@v1
+      - name: Get zizmor
+        run: cargo install zizmor
+      - name: Run zizmor 🌈
```
NoahStapp, line 24: Same here.

NoahStapp (changes requested):

.github/workflows/zizmor.yml
blink1073: No it was part of the example in the zizmore repo

NoahStapp: Then can we remove it? It's a little distracting to have emojis in config files.

blink1073: Done

NoahStapp (approved):

PR #313, opened 2024-11-12
MOTOR-1413 Install pymongo from v4.9 branch when installing from source

blink1073:


aclark4life (approved):
LGTM! 4.9 from source because … that's the newest that works with Motor?

blink1073:
> that's the newest that works with Motor?

Yes, we've pinned to that minor version.

PR #316, opened 2025-01-11
 MOTOR-1420 Add detailed return type to to_list method

SaidBySolo:
My type checker complains that there is an unknown type in the list.
I opened this PR because it seemed better to return Any rather than Unknown.


blink1073 (approved):
Thank you!

SaidBySolo:
Thank you for maintaining library! :D

PR #318, opened 2025-01-13
MOTOR-1420 Use generic return type for to_list()

ShaneHarvey:
MOTOR-1420 Use generic return type for to_list()

Small follow up for https://github.com/mongodb/motor/pull/316


blink1073 (approved):
LGTM!

PR #320, opened 2025-01-24
In documentation clarify the purpose behind client creation.

nfsknight:
In the documentation, I added a sentence in the asyncio tutorial clarifying the purpose behind client creation.


doc/tutorial-asyncio.rst
```diff
@@ -78,7 +78,7 @@ Motor, like PyMongo, represents data with a 4-level object hierarchy:
 
 Creating a Client
 -----------------
-Creating a client is what establishes a connection to MongoDB and tells your app what deployment (i.e. cluster) to connect to.  
+Creating a client is what establishes a connection to MongoDB and tells your app what deployment (i.e. cluster) to connect to.
```
ShaneHarvey, line 81: Is this needed? We already say above:
```
* :class:`~motor.motor_asyncio.AsyncIOMotorClient`
  represents a mongod process, or a cluster of them. You explicitly create one
  of these client objects, connect it to a running mongod or mongods, and
  use it for the lifetime of your application.
```

nfsknight: I think it would be helpful to less experienced developers who aren't familiar with the more technical terminology used in the definition under "Object Hierarchy" to begin by restating the purpose of creating a client in laymen's terms under "Creating a Client."

ShaneHarvey: I'm ambivalent on this change because it seems redundant but if we want to add it we should use the same phrase in the doc/tutorial-tornado.rst file.

nfsknight:
I added the phrase to the doc/tutorial-tornado.rst file as well.  I don't know what the evergreen patch thing means.

ShaneHarvey (approved):

PR #344, opened 2025-07-23
MOTOR-1473 Add support for Python 3.14 and drop 3.9 support

blink1073:


doc/changelog.rst
```diff
@@ -3,6 +3,11 @@ Changelog
 
 .. currentmodule:: motor.motor_tornado
 
+Motor 3.8.0
```
NoahStapp, line 6: Should we add the same warning we have below again to each release going forward until we fully deprecate Motor?

NoahStapp (changes requested):

doc/changelog.rst
blink1073: Done

NoahStapp (approved):

PR #386, opened 2026-06-01
MOTOR-1479 & MOTOR-1480 Fix build failures from PyMongo 4.17 changes

blink1073:
### Summary

Fixes build failures introduced by PyMongo 4.17:

- **MOTOR-1479**: PyMongo 4.17 (PYTHON-4542) added a `bind()` method to `ClientSession`. Motor is in maintenance mode and not adding new features, so `bind` is excluded from the session attribute parity check (matching how other PyMongo-only attributes are handled, e.g. `pymongo_client_only`).

- **MOTOR-1480**: PYTHON-5774 added `test_daemon.py` to PyMongo's test suite, which imports `pymongo.daemon`. The synchro test runner replaces the `pymongo` namespace with `synchro`, so `import pymongo.daemon` fails unless `daemon` is re-exported from `synchro/__init__.py`.

Additional fixes for failures exposed on PyMongo 4.17:

- **Typing**: Guard `time.mktime(date_tuple)` in `motor/web.py` against `None`, which `email.utils.parsedate` can return for malformed `If-Modified-Since` headers.

- **Exhaust cursor tests**: Skip `MotorClientExhaustCursorTest` — mockupdb's `OpMsg` support is insufficient for the current PyMongo wire protocol.

- **Linkcheck**: Fix broken external links (`#cursor-batches` and `#std-label-pymongo-mongodb-aws` anchors). Set `linkcheck_workers = 1` and `linkcheck_timeout = 60` per PYTHON-5847 to prevent the dochub redirect service from dropping connections under concurrent load.

### DIFF
--- a/motor/web.py
+++ b/motor/web.py
@@ -140,17 +140,17 @@ class GridFSHandler(tornado.web.RequestHandler):
         ims_value = self.request.headers.get("If-Modified-Since")
         if ims_value is not None:
             date_tuple = email.utils.parsedate(ims_value)
-            assert date_tuple is not None
 
-            # If our MotorClient is tz-aware, assume the naive ims_value is in
-            # its time zone.
-            if_since = datetime.datetime.fromtimestamp(time.mktime(date_tuple)).replace(
-                tzinfo=modified.tzinfo
-            )
+            if date_tuple is not None:
+                # If our MotorClient is tz-aware, assume the naive ims_value is in
+                # its time zone.
+                if_since = datetime.datetime.fromtimestamp(time.mktime(date_tuple)).replace(
+                    tzinfo=modified.tzinfo
+                )
 
-            if if_since >= modified:
-                self.set_status(304)
-                return
+                if if_since >= modified:
+                    self.set_status(304)
+                    return
 
         # Same for Etag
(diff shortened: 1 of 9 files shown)


motor/web.py
```diff
@@ -140,17 +140,17 @@ async def get(self, path, include_body=True):
...
         if ims_value is not None:
             date_tuple = email.utils.parsedate(ims_value)
-            assert date_tuple is not None
 
-            # If our MotorClient is tz-aware, assume the naive ims_value is in
-            # its time zone.
-            if_since = datetime.datetime.fromtimestamp(time.mktime(date_tuple)).replace(
-                tzinfo=modified.tzinfo
-            )
+            if date_tuple is not None:
```
NoahStapp, line 144: Intentional change over the `assert`?

NoahStapp (changes requested):

motor/web.py
blink1073: Yes because it actually can be `None`, which caused the build failure in my first commit.

blink1073:
Putting it in draft until evergreen passes

NoahStapp (approved):

PR #411, opened 2026-09-16
PYTHON-6075 Pin exact drivers-github-tools versions ahead of v4

sleepyStick:
https://jira.mongodb.org/browse/PYTHON-6075


sleepyStick:
i think the typing needs to be fixed upstream in pymongo? (I'm guessing its a result of the refactoring work)

.github/zizmor.yml
```diff
@@ -3,5 +3,5 @@ rules:
     config:
       policies:
         actions/*: ref-pin
-        mongodb-labs/drivers-github-tools/*: ref-pin
+        mongodb-labs/drivers-github-tools/*: hash-pin
```
blink1073: this is the default, so we can leave it out

blink1073:
> i think the typing needs to be fixed upstream in pymongo? (I'm guessing its a result of the refactoring work)

Yeah this PR should only affect the GitHub tests

.github/zizmor.yml
sleepyStick: ah didn't know that, removed!

blink1073 (approved):
LGTM!

PR #413, opened 2026-09-17
MOTOR-1482 fix typing failures

sleepyStick:


motor/core.pyi
```diff
@@ -35,13 +35,13 @@ from bson.raw_bson import RawBSONDocument
 from pymongo import IndexModel, ReadPreference, WriteConcern
 from pymongo.change_stream import ChangeStream
 from pymongo.client_options import ClientOptions
-from pymongo.client_session import ClientSession, SessionOptions, TransactionOptions
+from pymongo.client_session import ClientSession, SessionOptions, TransactionOptions  # type: ignore[attr-defined,unused-ignore]
```
sleepyStick: the code functionally works, but fails the typing tests because mypy's `strict=True` doesn't like when something is in the file via a `*` import. (which is how `SessionOptions` is imported into pymongo.client_session)

(the unused-ignore is cuz on different versions of pymongo, this typing error doesn't exist)

blink1073 (approved):
LGTM!

